PR #6, opened 2021-07-20
Compute md5sum of submitted bundle

Allda:
Content of all files related to operator bundle is hashed using mf5sum.
The sum is used to identity if content is still the same as was
submitted together with artifacts and test results.

JIRA: ISV-905


MarcinGinszt:
LGTM

PR #7, opened 2021-07-20
Hosted pipeline scaffolding

MarcinGinszt:
Scafflding for the Hosted pipeline- definitons of actual tasks, but with mock steps. To be filled with the next PRs.
Jira ticket: ISV-895


README.md
```diff
@@ -28,17 +31,50 @@ EOF
...
+curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/yaml-lint/0.1/yaml-lint.yaml | oc apply -f -
+
+
 tkn pipeline start operator-ci-pipeline \
   --param git_repo_url=https://github.com/Allda/operator-test-repo.git \
   --param git_repo_name=Allda/operator-test-repo \
   --param git_revision=master \
   --param bundle_path=operators/kogito-operator/1.6.0 \
+  --workspace name=pipeline,volumeClaimTemplateFile=templates/workspace-template.yml \
   --workspace name=pipeline,volumeClaimTemplateFile=test/workspace-template.yml \
```
Allda, line 49: You should remove this line since you added a new location for a template.

pipelines/operator-hosted-pipeline.yml
```diff
@@ -0,0 +1,255 @@
...
+    - name: git_repo_url
+    - name: git_username
+    - name: pyxis_url
+    - name: bundle_path
+    - name: preflight_latest_version
+    - name: ci_latest_version
+    - name: internal_registry_url
+      default: default-route-openshift-image-registry.apps-crc.testing
+    - name: quay_url
+      default: quay.io
```
amisstea, line 20: Can these be combined into a single param (e.g. `registry_url`)? We shouldn't be restricted to an internal registry and we can push everything to Quay.

pipelines/operator-hosted-pipeline.yml
```diff
@@ -0,0 +1,255 @@
...
+  params:
+    - name: git_pr_branch
+    - name: git_pr_title
+    - name: git_pr_url
+    - name: git_repo_url
+    - name: git_username
+    - name: pyxis_url
+    - name: bundle_path
+    - name: preflight_latest_version
+    - name: ci_latest_version
```
amisstea, line 16: Shouldn't 'latest' really be 'min' instead? These will be used to verify that the testing was performed using a supported set of tools, right?

pipelines/operator-hosted-pipeline.yml
```diff
@@ -0,0 +1,255 @@
...
+          subPath: src
+
+    - name: yaml-lint
+      runAfter:
+        - reserve-operator-name
+      taskRef:
+        name: yaml-lint
+      params:
+        - name: args
+          value: ["-d {extends: relaxed, rules: {line-length: {max: 180}}}", "$(params.bundle_path)"]
```
Allda, line 89: Let's use the same rules as in CI pipeline: https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/pipelines/operator-ci-pipeline.yml#L67

Relaxed profile validation is very limited and many things pass.

pipelines/operator-hosted-pipeline.yml
```diff
@@ -0,0 +1,255 @@
...
+      runAfter:
+        - reserve-operator-name
+      taskRef:
+        name: query-publishing-checklist
+      params:
+        - name: cert_project_id
+          value: "$(tasks.submission-validation.results.cert_project_id)"
+
+    # Try to retrieve results
+    - name: get-ci-results1
```
Allda, line 153: Let's find a better name for the task without a sequence number as a suffix.

pipelines/operator-hosted-pipeline.yml
```diff
@@ -0,0 +1,255 @@
...
+    - name: placeholder-run-ci-pipeline
+      runAfter:
+        - get-ci-results1
+      taskRef:
+        name: placeholder-run-ci-pipeline
+      params:
+        - name: result_resp_code
+          value: "$(tasks.get-ci-results1.results.result_resp_code)"
+
+    - name: get-ci-results2
```
Allda, line 188: Same here.

tasks/verify-changed-directories.yml
```diff
@@ -0,0 +1,19 @@
...
+  params:
+    - name: operator_name
+    - name: operator_version
+    - name: git_pr_branch
+  steps:
+    - name: get-cert-project-id
+      image: registry.access.redhat.com/ubi8-minimal
+      script: |
+        #! /usr/bin/env bash
+        # curl https://api.github.com/repos/operator-framework/community-operators/compare/master...$(params.git_pr_branch)
```
Allda, line 16: If I remember correctly GitHub APIs are very limited and after few queries, it stops responding due to free limits. I guess there should be a way how to get a diff using just `git` cli

MarcinGinszt: "For API requests using Basic Authentication or OAuth, you can make up to 5,000 requests per hour." 
https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting
I'm open to discussing if we should be careful with that

pipelines/operator-hosted-pipeline.yml
MarcinGinszt: get-ci-results-attempt and get-ci-results?

Allda (approved):

PR #8, opened 2021-07-21
Run preflight and build images in Hosted pipeline

MarcinGinszt:
The hosted pipeline has to repeat two functionalities after the CI pipeline:
1. Run the preflight tests (conditionally)
2. Build the bundle/ index images- and push them to quay.io instead of the default, internal registry.

For that, we should reuse existing tasks, after adding the parameters allowing necessary customization.
Jira task- ISV-904


Allda (approved):
LGTM

PR #11, opened 2021-07-26
Allow private registries

Allda:
The CI pipeline is not capable to push to a private registry using
provided credentials.

The original buildah task doesn't support a custom registry credentials
stored as "dockerconfigjson" so this commit adds a custom temporary
buildah task that use a secret workspace with token.
    
JIRA: ISV-855

This PR also replace git-clone ClusterTask with a Task.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,9 @@ To trigger a CI pipeline follow steps below:
 oc apply -R -f pipelines/operator-ci-pipeline.yml
 oc apply -R -f tasks
 
-# Install external yaml-lint task
+# Install external dependencies
 curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/yaml-lint/0.1/yaml-lint.yaml | oc apply -f -
+curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.4/git-clone.yaml | oc apply -f -
 
 
 tkn pipeline start operator-ci-pipeline \
@@ -83,8 +84,9 @@ To trigger a Hosted pipeline follow steps below:
 oc apply -R -f pipelines/operator-hosted-pipeline.yml
 oc apply -R -f tasks
 
-# Install external yaml-lint task
+# Install external dependencies
 curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/yaml-lint/0.1/yaml-lint.yaml | oc apply -f -
+curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.4/git-clone.yaml | oc apply -f -
 
 tkn pipeline start operator-hosted-pipeline \
   --param git_pr_branch=main \
--- a/README.md
+++ b/README.md
@@ -31,6 +31,26 @@ EOF
 oc create -f ssh-secret.yml
 ```
 
+The CI pipeline supports a custom private registries where bundle and test
+index are pushed. To allow a private registry access user need to create
+auth secret - .dockerconfigjson with registry username and access token.
+
+```bash
+# Download an access token and create a secret
+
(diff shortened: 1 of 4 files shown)


tasks/buildah.yml
```diff
@@ -0,0 +1,85 @@
+---
```
MarcinGinszt, line 1: Why it is so scrambled in comparison with the original task? It is hard to compare it with the original task so that somebody in the future could have problems understanding what was changed here.
Could you also please add some small note, mentioning the source of the original task (https://github.com/tektoncd/catalog/blob/main/task/buildah/0.2/buildah.yaml) and what was changed there?

README.md
```diff
@@ -31,22 +31,46 @@ EOF
...
+# Install external dependencies
 curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/yaml-lint/0.1/yaml-lint.yaml | oc apply -f -
+curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.4/git-clone.yaml | oc apply -f -
 
 
 tkn pipeline start operator-ci-pipeline \
-  --param git_repo_url=https://github.com/Allda/operator-test-repo.git \
-  --param git_repo_name=Allda/operator-test-repo \
-  --param git_revision=master \
+  --param git_repo_url=git@github.com:redhat-openshift-ecosystem/operator-pipelines-test-repo.git \
```
MarcinGinszt, line 65: I think you should do that for both of the pipelines

MarcinGinszt (changes requested):
Apart from two comments- LGTM, good job!

tasks/buildah.yml
Allda: The buildah task versions that are available in tekton hub don't work on our clusters because it requires a privileged access. That is why I took the buildah task directly from `ClusterTask` from a cluster and just replaced `ClusterTask` with `Task`

README.md
Allda: Added

README.md
```diff
@@ -31,22 +31,46 @@ EOF
...
 ```
 
+The CI pipeline supports a custom private registries where bundle and test
+index are pushed. To allow a private registry access user need to create
+auth secret - .dockerconfigjson with registry username and access token.
+
+```bash
+# Download an access token and create a secret
+
+cat << EOF > redhat-isv-redhat-isv-robot-secret.yml
```
amisstea, line 41: Do we really need to repeat "redhat-isv" twice here and elsewhere?

Allda: The file was generated by quay.io when generating a token and this is the only example of how to use it so I guess it is no big deal.

tasks/buildah.yml
MarcinGinszt: Still, just a small part of the logic is changed compared to the tektonhub task,  but a large part of the definition is changed (eg. in array "params"
Original:
```
  params:
  - name: IMAGE
    description: Reference of the image buildah will produce.
  ```
  
Our definition:
  ```
      - description: Reference of the image buildah will produce.
      name: IMAGE
      type: string
```

Allda:
@MarcinGinszt I addressed your comments. Can you check it again? Thanks

README.md
MarcinGinszt: Thank you!

MarcinGinszt (changes requested):
What do you think about adding just necessary changes to the buildah task and documenting them with a comment?

tasks/buildah.yml
MarcinGinszt: After discussion with @Allda, he will create a task to push the changes in the Buildah task definition in tekton hub, and then we'll replace the task definition

MarcinGinszt (approved):
LGTM

PR #13, opened 2021-07-27
Update pin manifest digest step

Allda:
The pinning step returns a dirty flag based on if there are any changes
in manifest references. The pinning step can be also skipped by setting
pipeline test_mode parameter.

JIRA: ISV-849

### DIFF
--- a/tasks/digest-pinning.yml
+++ b/tasks/digest-pinning.yml
@@ -6,6 +6,7 @@ metadata:
 spec:
   params:
     - name: bundle_path
+    - name: skip
   results:
     - name: dirty_flag
   workspaces:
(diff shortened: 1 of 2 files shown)


Allda:
@amisstea @MarcinGinszt Please review

amisstea (approved):
LGTM :+1:

tasks/digest-pinning.yml
```diff
@@ -18,8 +19,20 @@ spec:
...
+          exit 0
+        fi
+
         BUNDLE_PATH=$(realpath $(params.bundle_path))
         ls -l $BUNDLE_PATH
         operator-manifest pin $BUNDLE_PATH
 
-        echo "true" | tee $(results.dirty_flag.path)
+        if [[ $(git diff --stat) != '' ]]; then
+          echo "Manifests were not pinned."
```
MarcinGinszt, line 33: When I tried this condition on my computer it doesn't work properly. 

I tried to run this script in the repository with no changes:
```
if [[ $(git diff --stat) != '' ]]; then
  echo "Manifests were not pinned."
else
  echo "Manifests are pinned."
fi
```
Result- "Manifests are pinned.".

MarcinGinszt:
LGTM

PR #14, opened 2021-07-28
Non-permissive additive index generation

amisstea:
The index generation task needs creds to pull from all private registries.
This requires the user to provide the terms-based registry creds
themselves rather than using the cluster-scoped secret because they
aren't accessible to OPM inside the container.


wcheang (approved):
I don't think I know enough to be able to spot anything wrong at this point, but it looks okay to me as far as my understanding goes :+1:

haripate:
Looks good to me!

haripate (approved):
Looks good to me!

MarcinGinszt:
LGTM

MarcinGinszt (approved):

PR #16, opened 2021-07-30
Upload preflight test artifacts and results

Allda:
The upload task reads a test results and artifacts and upload them using
Pyxis API to RH storage. This is later used by hosted pipeline to skip
preflight tests.

JIRA: ISV-883


MarcinGinszt (approved):
LGTM.
I'll add the `-n` flag to echo in hosted pipeline tasks.

PR #17, opened 2021-08-02
Submission validation script

MarcinGinszt:
ISV-906
Related PR: https://github.com/redhat-openshift-ecosystem/operator-pipelines-images/pull/9
- Distinguished script `certification-project-check`
- Used script `verify-submission`

### DIFF
--- a/tasks/get-ci-results.yml
+++ b/tasks/get-ci-results.yml
@@ -20,7 +20,7 @@ spec:
         #! /usr/bin/env bash
         PREFLIGHT_RESULTS_EXISTS="$(params.preflight_results_exists)"
         if [ $PREFLIGHT_RESULTS_EXISTS == "true" ]; then
-          echo "Preflight results already exists- skipping"
+          echo -n "Preflight results already exists- skipping"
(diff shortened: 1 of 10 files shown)


tasks/get-ci-results.yml
```diff
@@ -20,7 +20,7 @@ spec:
         #! /usr/bin/env bash
         PREFLIGHT_RESULTS_EXISTS="$(params.preflight_results_exists)"
         if [ $PREFLIGHT_RESULTS_EXISTS == "true" ]; then
-          echo "Preflight results already exists- skipping"
+          echo -n "Preflight results already exists- skipping"
```
Allda, line 23: `-n` option is useful only when returning a value as a result. For classic log messages, the option doesn't have to be there.

MarcinGinszt: Ah, yes. Sorry, `find and replace` went out of control- thanks for your catch!

wcheang (approved):

Allda (approved):

PR #18, opened 2021-08-02
Make use of imagestreams to pull index images

amisstea:
This reduces the complexity for the user to get started because they no
longer need to specify credentials for the internal registry or
terms-based registry. The trade-off is the addition of two one-time
configuration commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ oc create -f registry-secret.yml
 
 The pipelines must pull the parent index images through the internal OpenShift
 registry to take advantage of the built-in credentials for Red Hat's terms-based
-registry (registry.redhat.io). The saves the user from needing to provide such
+registry (registry.redhat.io). This saves the user from needing to provide such
 credentials. The index generation task will always pull published index images
 through imagestreams of the same name in the current namespace. As a result,
 there is a one time configuration for each desired distribution catalog.
@@ -99,7 +99,18 @@ curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.4/
 ```
 
 ### Execution
-The CI pipeline can be triggered using the tkn CLI like so:
+If using the default internal registry, the CI pipeline can be triggered using the tkn CLI like so:
+
+```bash
+tkn pipeline start operator-ci-pipeline \
+  --param git_repo_url=git@github.com:redhat-openshift-ecosystem/operator-pipelines-test.git \
+  --param git_revision=main \
+  --param bundle_path=operators/kogito-operator/1.6.0-ok \
+  --workspace name=pipeline,volumeClaimTemplateFile=templates/workspace-template.yml \
+  --workspace name=ssh-dir,secret=my-ssh-credentials \
+  --showlog
+```
+If using an external registry, the CI pipeline can be triggered using the tkn CLI like so:
 
 ```bash
 tkn pipeline start operator-ci-pipeline \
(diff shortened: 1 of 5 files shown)


tasks/generate-index.yml
```diff
@@ -24,13 +28,23 @@ spec:
...
+            $HOME/.docker/config.json \
+            > $DOCKER_CONFIG/config.json
+        fi
+
+        INDEX_IMG_STREAM=$(echo $(params.from_index) | rev | cut -f1 -d/ | rev)
+        FROM_INDEX="$(params.internal_registry)/$(context.taskRun.namespace)/$INDEX_IMG_STREAM"
 
         opm index add \
-          --from-index "$(params.from_index)" \
+          --from-index $FROM_INDEX \
```
MarcinGinszt, line 47: What if it's a new operator? We'll get the error, that such an image doesn't exist, right?

README.md
```diff
@@ -57,6 +53,30 @@ EOF
 oc create -f my-registry-secret.yml
 ```
 
+#### Red Hat Catalog Imagestreams
+
+The pipelines must pull the parent index images through the internal OpenShift
+registry to take advantage of the built-in credentials for Red Hat's terms-based
+registry. There is a one time configuration for each desired distribution catalog.
```
MarcinGinszt, line 60: That is just my opinion that you can disagree with, but I think it's hard to understand the purpose of following actions from the first sentence.

README.md
```diff
@@ -57,6 +53,30 @@ EOF
...
+
+```bash
+# Must be run once before certifying against the certified catalog.
```
MarcinGinszt, line 63: Why only once, since it imports the latest information? What if, meanwhile, somebody will push a newer version of the image, and the information will change?

README.md
```diff
@@ -34,14 +34,10 @@ oc create -f ssh-secret.yml
...
 #### Registry Credentials
-The CI pipeline requires credentials to push and/or pull from all authenticated
-registries. This includes:
-
-* Red Hat terms-based registry (registry.redhat.io)
-* The registry specified by the `registry` param (including the internal Openshift registry)
-
-The user must create an auth secret containing the docker config with all
-credentials included. For example:
+The CI pipeline can optionally be configured to push images to a remote private
```
MarcinGinszt, line 37: How can user not specify the credentials?
After I removed the parameters `registry` and workspace `credentials`, I've got the error 
```
[build-bundle : push] + buildah --storage-driver=vfs push --tls-verify=true --digestfile /workspace/source/image-digest image-registry.openshift-image-registry.svc:5000/redhat-isv/kogito-operator:1.6.0-ok docker://image-registry.openshift-image-registry.svc:5000/redhat-isv/kogito-operator:1.6.0-ok
[build-bundle : push] Getting image source signatures
[build-bundle : push] unauthorized: authentication required
[build-bundle : push] level=error msg="exit status 125"
```

Allda (approved):

README.md
amisstea: Are you specifying the `image_stream` param as well? It seems to be poorly named but it's the namespace the image gets pushed. You can remove that param and it should default to the current pipeline namespace.

MarcinGinszt: Could you please add one sentence, explaining how the parameters are different for pipeline pointing to the private registry?

README.md
amisstea: I added some more detail. Does that help?

README.md
amisstea: I added an example invocation for the internal registry and renamed the param.

MarcinGinszt: Thank you, that's great

README.md
MarcinGinszt: Yes, thank you! That's great. There is just a small typo there: 
`The saves the user`

MarcinGinszt:
One typo in comment and LGTM!

README.md
amisstea: Fixed.

PR #20, opened 2021-08-03
Add deployment playbooks for hosted pipeline

Allda:
This commit contains an Ansible configuration that deploys a hosted
pipeline to OpenShift environment. The new role uses k8s module to
communicate with OCP.

JIRA: ISV-918


Allda:
@amisstea @MarcinGinszt Feel free to review the PR. The prod ansible config is not final because I still don't have a cluster with a service account.

ansible/init.sh
```diff
@@ -0,0 +1,77 @@
...
+        echo "dev stage qa prod"
+    else
+        echo "$PARAMS"
+    fi
+}
+
+# execute playbook for given environment
+execute_playbook() {
+    local secret=$(dirname "$0")/vaults/$env/secret-vars.yml
+    if [ ! -z $secret ]; then
```
MarcinGinszt, line 29: You can use -n instead of -z:
```
-n STRING | The length of STRING is greater than zero.
-z STRING | The lengh of STRING is zero (ie it is empty).
```
shellcheck told me that, I recommend this tool

MarcinGinszt: Btw. What this condition checks? This string is just concatenated, it never has length equal to 0, script always executes
`echo "File $secret was not found, empty one was created"`

MarcinGinszt:
I don't know the Ansible, so I couldn't really make a quality code review

ansible/init.sh
Allda: Good point, I originally copied the script from our other services. I changed it to -f which should be better here.

ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+  k8s:
+    state: present
+    force: yes
+    namespace: "{{ oc_namespace }}"
+    definition:
+      apiVersion: v1
+      kind: Secret
+      type: Opaque
+      metadata:
+        name: operator-pipeline-secrets-{{ suffix }}
```
MarcinGinszt, line 43: What do you think about changing this secret name to something more meaningful, eg. pyxis-auth-cert?

ansible/roles/operator-pipeline/templates/openshift/tasks/git-clone.yml
```diff
@@ -0,0 +1,213 @@
+apiVersion: tekton.dev/v1beta1
```
MarcinGinszt, line 1: Why aren't we using the ready-made task here?

Allda: I tried it in the first version, but k8s can't deal with remote resources so I pull the task into repo directly.

ansible/roles/operator-pipeline/tasks/main.yml
Allda: I defined the name the same way as we do it with any other service. Usually, we have one openshift secret with multiple files so it is easier when mounting all secrets to DC/pod/pipeline.

haripate (approved):
LGTM

PR #21, opened 2021-08-09
Using the script to download artifacts

MarcinGinszt:
Related PR: https://github.com/redhat-openshift-ecosystem/operator-pipelines-images/pull/10
Jira: ISV-910

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results-attempt.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results-attempt.yml
@@ -23,24 +23,26 @@ spec:
     - name: results
     - name: pyxis-ssl-credentials
       optional: true
-  env:
-    - name: PYXIS_CERT_PATH
-      value: $(params.pyxis_cert_path)
-    - name: PYXIS_KEY_PATH
-      value: $(params.pyxis_key_path)
   steps:
     - name: get-results
       image: quay.io/redhat-isv/operator-pipelines-images:latest
       workingDir: $(workspaces.results.path)
+      env:
+        - name: PYXIS_CERT_PATH
+          value: $(params.pyxis_cert_path)
+        - name: PYXIS_KEY_PATH
+          value: $(params.pyxis_key_path)
       script: |
         #! /usr/bin/env bash
+
+        set -xe
 
         # Send request to Pyxis to get the logs and test results.
         # If they exists, store them in the workspace
 
         echo "Downloading the artifacts"
 
-        download-artifacts \
+        download-test-results \
           --pyxis-url "$(params.pyxis_url)" \
           --cert-project-id "$(params.cert_project_id)" \
           --certification-hash "$(params.md5sum)" \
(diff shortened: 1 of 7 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results.yml
```diff
@@ -10,18 +10,61 @@ spec:
...
+        #! /usr/bin/env bash
+
+        # Send request to Pyxis to get the logs and test results.
+        # If they exists, store them in the workspace
+
+        echo "Downloading the artifacts"
+
+        download-artifacts \
+          --pyxis-url "$(params.pyxis_url)" \
+          --pyxis-url https://pyxis.stage.engineering.redhat.com
```
wcheang, line 52: The pyxis-url input is duplicated. Is this a mistake?

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results-attempt.yml
```diff
@@ -9,16 +9,59 @@ spec:
...
+
+        echo "Downloading the artifacts"
+
+        download-artifacts \
+          --pyxis-url "$(params.pyxis_url)" \
+          --cert-project-id "$(params.cert_project_id)" \
+          --certification-hash "$(params.md5sum)" \
+          --operator-package-version "$(params.bundle_version)" \
+          --operator-name "$(params.operator_name)" \
+          --cert-path /cert/tls.crt \
```
wcheang, line 37: There will be multiple different tasks that need the cert and key. The paths should probably be defined at the pipeline level then passed into the task, instead of hardcoding it in each task.

wcheang (commented):
I know this was not part of the changes in this PR, but can we rename get-ci-results-attempt to something more descriptive, maybe by adding "preflight" to it? Just looking at the names I couldn't tell why there are get-ci-results and get-ci-results-attempts, until I saw the comment on the pipeline file.

ansible/roles/operator-pipeline/templates/openshift/tasks/verify-pinned-digest.yml
```diff
@@ -15,5 +15,5 @@ spec:
         echo $DIRTY_FLAG
         if [ $DIRTY_FLAG == "true" ]; then
           echo "There are unpinned images digests!"
-          exit 1
+          #exit 1
```
wcheang, line 18: Probably want to uncomment this?

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results-attempt.yml
```diff
@@ -9,16 +9,59 @@ spec:
     - name: pyxis_url
     - name: cert_project_id
     - name: bundle_version
+    - name: operator_name
   results:
     - name: preflight_results_exists
+    - name: test_result_id
+    - name: test_logs_id
```
amisstea, line 16: Why do we need to get the log? I don't believe we need this in the hosted pipeline.

wcheang (approved):

Allda (approved):

PR #22, opened 2021-08-11
Add call to script for reserving operator name

wcheang:


wcheang:
@MarcinGinszt @amisstea this still needs to be tested and have its conflicts resolved, but would you please take a quick look to see if I'm on the right track? This goes with https://github.com/redhat-openshift-ecosystem/operator-pipelines-images/pull/12

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
```diff
@@ -30,4 +30,12 @@ spec:
           exit 1
         fi
 
+        ISV_PID=$(cat $CI_FILE_PATH | yq -r '.isv_pid')
```
amisstea, line 33: We should get the `isv_pid` from their certProject. We already have the project id to do that lookup.

amisstea (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
wcheang: Okay, so make another request in the reserve-operator-name script to GET the certProject data?

amisstea: A better thing would be to retrieve the certProject record, even if just via curl, in this step and store it in a workspace. Then it can be referenced again in followup steps and tasks.

ansible/roles/operator-pipeline/templates/openshift/tasks/reserve-operator-name.yml
```diff
@@ -5,11 +5,37 @@ metadata:
...
+          --operator-package-name "$(params.operator_name)" \
+          --source "certified-operators"  \
+          --verbose
+
+        if [ $? -eq 0 ]; then
+          echo "Operator name reserved successfully"
+        else
+          echo "Failed to reserve operator name. Name may have already been taken."
+          exit 1
+        fi
```
amisstea, line 32: I noticed that `reserve-operator-name` already returns a non-0 exit code on error with descriptive messages. So we can probably remove these lines and add `set -e` at the start of this script to exit automatically.

amisstea:
@wcheang is this ready for merging?

wcheang:
> @wcheang is this ready for merging?

@amisstea I still haven't been able to run this on OSD as part of the whole pipeline yet, but I've tested commands individually. If that's okay then it's ready for merging. Otherwise still working on it :sweat_smile:

amisstea:
No worries! It just looked like it was nearing completion. Please do test with it integrated into the pipeline though.

wcheang:
Turns out the REST calls to Pyxis in the pipeline needs CA certs imported to prevent SSL errors. This was done by @Allda in https://github.com/redhat-openshift-ecosystem/operator-pipelines-images/pull/14

There is also some conflicting implementation of getting field(s) from cert project...it'll probably be easier for everyone involved if the approach is finalized in Ales's PR, and I'll just adjust this one accordingly.

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
```diff
@@ -30,4 +41,15 @@ spec:
           exit 1
         fi
 
+        PROJECT_URL="$(params.pyxis_url)/v1/projects/certification/id/$CERT_PROJECT_ID"
+        PROJECT=$(curl -k -X GET $PROJECT_URL -H  "accept: application/json" --cert $(params.pyxis_cert_path) --key $(params.pyxis_key_path))
```
wcheang, line 45: I discussed with Ales a bit about using a Python script to get the cert project and then extract all the needed fields, but that'll require refactoring operator-pipeline-images and will further delay this. I'd rather merge this first then refactor.

wcheang:
@amisstea @MarcinGinszt @haripate @Allda this is ready for review now. Please take a look.

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
```diff
@@ -30,4 +41,15 @@ spec:
           exit 1
         fi
 
+        PROJECT_URL="$(params.pyxis_url)/v1/projects/certification/id/$CERT_PROJECT_ID"
```
Allda, line 44: Pipeline parameter pyxis_url uses a trailing `/` so I am not really sure if this concatenation would work.

ansible/roles/operator-pipeline/templates/openshift/tasks/reserve-operator-name.yml
```diff
@@ -5,11 +5,35 @@ metadata:
   name: reserve-operator-name
 spec:
   params:
+    - name: association
     - name: operator_name
+    - name: pyxis_cert_path
+      default: ""
+      description: Path to Pyxis certificates. Valid only when internal Pyxis is used.
```
Allda, line 12: Can you add a description to all params?

Allda (commented):
Few minor comments, but concept looks good

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
```diff
@@ -30,4 +41,15 @@ spec:
...
+
+        ISV_PID=$(echo $PROJECT | jq -r '.container.isv_pid')
+
+        if [ -z $ISV_PID ]; then
+          echo "Certification project isv_pid not found."
+          exit 1
+        fi
+
         echo -n $CERT_PROJECT_ID | tee $(results.certification_project_id.path)
+        echo -n $ISV_PID | tee $(results.isv_pid.path)
```
MarcinGinszt, line 55: Why do we need isv_pid?

wcheang: isv_pid is inputted into the reserve-operator-name task as the "assocation"

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
wcheang: Just tested on my command line, it should work with the double `/`. I thought adding the `/` is safer in case the input doesn't contain a trailing one.

PR #23, opened 2021-08-11
Release pipeline scaffolding

MarcinGinszt:
Jira task: ISV-761


Allda (changes requested):
Looks good to me. There are some yamlint failures that need to be fixed first. The release pipeline should be also released to openshift so it would be great to add the pipeline to the release step:

https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/tasks/main.yml#L68

README.md
```diff
@@ -165,3 +166,38 @@ tkn pipeline start operator-hosted-pipeline \
...
+
+### Execution
+The release pipeline can be triggered using the tkn CLI like so:
+
+
+tkn pipeline start operator-release-pipeline \
+  --param git_repo_url=git@github.com:redhat-openshift-ecosystem/operator-pipelines-test.git \
+  --param git_revision=main \
+  --param bundle_path=operators/kogito-operator/1.6.0-ok \
+  --param container_digest=sha256:22e172e14b1300ce473ca15a80c273676305ad6c7a57f0f8cb25816bfce8d196 \
```
MarcinGinszt, line 195: @amisstea I'm confused, where this should come from. Are we missing step to compute that? How does it already exists in the Periwinkle?

ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -77,11 +77,20 @@
...
+- name: Deploy hosted pipeline
   k8s:
     state: present
     apply: yes
     namespace: "{{ oc_namespace }}"
     definition: "{{ lookup('template', '{{ item }}') }}"
   with_items:
     - ../templates/openshift/pipelines/operator-hosted-pipeline.yml
+
+- name: Deploy release pipeline
```
Allda, line 89: You can deduplicate few lines of code by adding a release pipeline to `with_item` section in the generic deploy pipeline task.

Allda (approved):

README.md
```diff
@@ -165,3 +166,38 @@ tkn pipeline start operator-hosted-pipeline \
   --workspace name=registry-credentials,secret=my-registry-secret \
   --showlog
 ```
+
+
+## Operator Release pipeline
+The Release pipeline runs after the layers of validation (CI (optionally) and Hosted pipeline).
+It is used to certify and publish submitted bundle version.
+It is triggered by merging the submission pull request and sucessfully completes with updating the GitHub
+indicating published versions.
```
amisstea, line 175: How about this instead?

> It is triggered by a merged pull request and successfully completes once the bundle has been distributed to all relevant catalogs and appears in the Red Hat Ecosystem Catalog.

It's probably more important to state that it succeeds when the distribution is complete. The github statuses are just a nice-to-have feature.

README.md
```diff
@@ -165,3 +166,38 @@ tkn pipeline start operator-hosted-pipeline \
...
+
+### Installation
+
+```bash
+oc apply -R -f ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
+oc apply -R -f ansible/roles/operator-pipeline/templates/openshift/tasks
+
+# Install external dependencies
+curl https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.4/git-clone.yaml | oc apply -f -
```
amisstea, line 184: We could simplify this command (and the others like it) like this:

> oc apply -f https://raw.githubusercontent.com/tektoncd/catalog/main/task/git-clone/0.4/git-clone.yaml

README.md
```diff
@@ -165,3 +166,38 @@ tkn pipeline start operator-hosted-pipeline \
...
+### Execution
+The release pipeline can be triggered using the tkn CLI like so:
+
+
+tkn pipeline start operator-release-pipeline \
+  --param git_repo_url=git@github.com:redhat-openshift-ecosystem/operator-pipelines-test.git \
+  --param git_revision=main \
+  --param bundle_path=operators/kogito-operator/1.6.0-ok \
+  --param imagestream=operator-pipelines \
+  --param tag=1.6.0-ok \
```
amisstea, line 196: Do we really need this param or could it be optional? By default I think it can come from the bundle.

README.md
```diff
@@ -165,3 +166,38 @@ tkn pipeline start operator-hosted-pipeline \
...
+  --param dist_method=marketplace \
```
amisstea, line 197: We can get this from the `certProject` so it doesn't need to be a param.

README.md
```diff
@@ -165,3 +166,38 @@ tkn pipeline start operator-hosted-pipeline \
...
+  --param pyxis_url=https://catalog.redhat.com/api/containers/ \
```
amisstea, line 198: We should use a default value for this param and also utilize the internal prod instance instead.

README.md
```diff
@@ -165,3 +166,38 @@ tkn pipeline start operator-hosted-pipeline \
...
+
+### Execution
+The release pipeline can be triggered using the tkn CLI like so:
+
+
+tkn pipeline start operator-release-pipeline \
+  --param git_repo_url=git@github.com:redhat-openshift-ecosystem/operator-pipelines-test.git \
+  --param git_revision=main \
+  --param bundle_path=operators/kogito-operator/1.6.0-ok \
+  --param imagestream=operator-pipelines \
```
amisstea, line 195: Is this the temporary bundle imagestream as pushed by the hosted pipeline (effectively a Quay namespace)? For the destination imagestream, that needs to come from the `containerVendor.label` field.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+    - name: tag
+    - name: bundle_path
+    - name: dist_method
+    - name: is_latest
+    - name: registry
+      default: quay.io
+    - name: image_namespace
+      default: $(context.pipelineRun.namespace)
+      description: The namespace/organization all built images will be pushed to.
+    - name: pyxis_url
```
amisstea, line 21: Please add descriptions for all of these and try to do the same in the tasks.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-bundle-related-data.yml
```diff
@@ -0,0 +1,22 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: get-bundle-related-data
```
amisstea, line 5: The name `get-bundle-related-data` is very generic and I so thought it would grab data from the bundle but that's not what it actually does. How about renaming this to `get-project-related-data` instead? Maybe it should even be merged with `certification-project-check` but I'll leave that for you to decide :smiley:

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+
+    # Publish Vendor, Repository
+    - name: publish-resources
+      runAfter:
+        - imagestream-manipulation
+      taskRef:
+        name: publish-resources
+      params:
+        - name: cert_project_id
+          value: "$(tasks.certification-project-check.results.certification_project_id)"
```
amisstea, line 139: We should probably publish the vendor and repo after the image is published. Given this order, if the repo gets published and then the pipeline fails, the repo will appear on the catalog search but the detail page may fail to load because of the lack of published image metadata.

ansible/roles/operator-pipeline/templates/openshift/tasks/publish-bundle.yml
```diff
@@ -0,0 +1,20 @@
...
+  params:
+    - name: cert_project_id
+    - name: container_digest
+    - name: bundle_versions
+  steps:
+    - name: publish-bundle
+      image: registry.access.redhat.com/ubi8-minimal
+      script: |
+        #! /usr/bin/env bash
+        echo "Starting bundle publish process"
```
amisstea, line 16: Please add a placeholder in here to set a _pending_ github status for each IIB job when it starts (in addition to when it completes). I'm not sure if you want to embed that in this task or make use of https://hub.tekton.dev/tekton/task/github-set-status so it's up to you for now.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -0,0 +1,178 @@
...
+    - name: is_latest
+    - name: registry
+      default: quay.io
+    - name: image_namespace
+      default: $(context.pipelineRun.namespace)
+      description: The namespace/organization all built images will be pushed to.
+    - name: pyxis_url
+  workspaces:
+    - name: repository
+    - name: ssh-dir
```
amisstea, line 24: Can we make this optional? It's only used to checkout the source which can be done with https.

amisstea (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/get-bundle-related-data.yml
MarcinGinszt: I'll pass merging it :slightly_smiling_face:  `certification-project-check` is used separately in other pipelines.

ansible/roles/operator-pipeline/templates/openshift/tasks/publish-bundle.yml
MarcinGinszt: I'd prefer to keep it in one task. Tekton task which you shared basically does one call to GitHub API
https://github.com/tektoncd/catalog/blob/main/task/github-set-status/0.2/github-set-status.yaml

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
MarcinGinszt: What can be the use case to trigger the internal pipelines different way than intended by design? 
Speaking of which, we would have to include other ways of authentication in docs.
Overall, even ssh-dir might not be necessary here, since we expect to check out only publicly visible repositories...

amisstea: It's very unlikely this will ever be used with private repos but leaving this optional will allow us to support that. Otherwise the repo will always be public with no need for write access.

amisstea (approved):
LGTM :+1:

PR #24, opened 2021-08-12
[ISV-850] The changes are pushed to temporary branch for pinned digest

haripate:
Here are the few temporary branches created for a testing purpose: https://github.com/haripate/operator-pipelines-test/branches


ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,99 @@ apiVersion: tekton.dev/v1beta1
 kind: Task
 metadata:
   name: commit-pinned-digest
+  labels:
```
MarcinGinszt, line 6: We are not using labels for our tasks yet. However it's interesting idea, let's keep it consistent with existing ones :slightly_smiling_face:

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,99 @@ apiVersion: tekton.dev/v1beta1
...
+        Absolute path to the user's home directory. Set this explicitly if you are running the image as a non-root user or have overridden
+        the gitInitImage param with an image containing custom user configuration.
+      type: string
+      default: "/tekton/home"
+    - name: VERBOSE
+      description: Log the commands that are executed during `git-clone`'s operation.
+      type: string
+      default: "true"
+  results:
+    - name: commit
```
MarcinGinszt, line 47: We don't use this result, it can be discarded

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,99 @@ apiVersion: tekton.dev/v1beta1
...
+          fi
+
+          if [ "${WORKSPACE_SSH_DIRECTORY_BOUND}" = "true" ] ; then
+            cp -R "${WORKSPACE_SSH_DIRECTORY_PATH}" "${PARAM_USER_HOME}"/.ssh
+            chmod 700 "${PARAM_USER_HOME}"/.ssh
+            chmod -R 400 "${PARAM_USER_HOME}"/.ssh/*
+          fi
+
+          # Setting up the config for the git.
+          git config --global user.email "$(params.GIT_USER_EMAIL)"
```
MarcinGinszt, line 82: We are not supplying the git credentials, we use SSH auth. Please, try to keep only necessary things in the script

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,99 @@ apiVersion: tekton.dev/v1beta1
...
+    - name: VERBOSE
+      description: Log the commands that are executed during `git-clone`'s operation.
+      type: string
+      default: "true"
+  results:
+    - name: commit
+      description: The precise commit SHA that was fetched by this Task.
   steps:
     - name: commit-changes
       image: registry.access.redhat.com/ubi8-minimal
```
MarcinGinszt, line 51: This image doesn't contain the "git" binary. Firstly, test the pipeline with the script and submit the PR when it will successfully create a PR to some repository. Thank you :slightly_smiling_face:

MarcinGinszt (changes requested):
That's a good start for the rest of the work, but you have to test the script and limit the functionalities to necessary ones

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,99 @@ apiVersion: tekton.dev/v1beta1
...
+            chmod -R 400 "${PARAM_USER_HOME}"/.ssh/*
+          fi
+
+          # Setting up the config for the git.
+          git config --global user.email "$(params.GIT_USER_EMAIL)"
+          git config --global user.name "$(params.GIT_USER_NAME)"
+
+          TEMP_BRANCH = "pinned-operator-manifest"
+          git checkout -b "${TEMP_BRANCH}"
+          git add --all
```
Allda, line 87: `--all` is a bit dangerous because there can be other files generated in previous steps. I think `-u` is a better option that only includes tracked files.

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
Allda: The new commit sha will be used when opening a new PR at the end of a pipeline so you can keep it.

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,99 @@ apiVersion: tekton.dev/v1beta1
...
+            cp -R "${WORKSPACE_SSH_DIRECTORY_PATH}" "${PARAM_USER_HOME}"/.ssh
+            chmod 700 "${PARAM_USER_HOME}"/.ssh
+            chmod -R 400 "${PARAM_USER_HOME}"/.ssh/*
+          fi
+
+          # Setting up the config for the git.
+          git config --global user.email "$(params.GIT_USER_EMAIL)"
+          git config --global user.name "$(params.GIT_USER_NAME)"
+
+          TEMP_BRANCH = "pinned-operator-manifest"
```
Allda, line 85: Let's add some suffix to a branch name (package + version?). Users can work on multiple parallel bundles so hardcoded name would make a conflict.

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
Allda: The open PR task actually needs a new branch name, can you please return the branch name as result as well?

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,93 @@ apiVersion: tekton.dev/v1beta1
...
+          git commit -m "Pin the operator manifests"
+          git push origin "${TEMP_BRANCH}"
+
+          RESULT_SHA="$(git rev-parse HEAD | tr -d '\n')"
+          # Make sure we don't add a trailing newline to the result!
+          echo -n "$RESULT_SHA" > $(results.commit.path)
+          echo -n "$TEMP_BRANCH" > $(results.branch.path)
+
         else
           echo "Skipping commit"
```
Allda, line 94: In case of the skipped commit, you should return some value in the result variables. Since the branch name is unknown you can set commit has to current sha and branch name set to empty string ""

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -3,21 +3,95 @@ apiVersion: tekton.dev/v1beta1
...
+        Git user email for performing git operation.
+      default: "no-reply@redhat.com"
+  results:
+    - name: commit
+      description: The precise commit SHA that was fetched by this Task.
+    - name: branch
   steps:
     - name: commit-changes
-      image: registry.access.redhat.com/ubi8-minimal
+      image: docker.io/alpine/git:v2.26.2@sha256:23618034b0be9205d9cc0846eb711b12ba4c9b468efdd8a59aac1d7b1a23363f #tag: v2.26.2
```
Allda, line 47: Any specific reason why to use an alpine image? And why there are both tags and digest?

amisstea: Good point. If it's possible to use ubi, we should.

Allda (approved):

PR #30, opened 2021-08-20
Add a CI pipeline param to control submission

amisstea:
The "when" condition is safe in this case because result/artifact
uploads and PR creation are the last steps.


MarcinGinszt (approved):
LGTM

PR #38, opened 2021-08-31
Add steps to set github status

wcheang:


ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
```diff
@@ -0,0 +1,83 @@
...
+
+        PREFLIGHT_RESULTS_EXISTS="$(params.preflight_results_exists)"
+        if [ $PREFLIGHT_RESULTS_EXISTS == "true" ]; then
+          echo "Preflight results already exists - skipping setting github status"
+          exit 0
+        fi
+
+        DESCRIPTION="$(params.description)"
+
+        if [ -f "$(params.result_file)" ]; then
```
MarcinGinszt, line 46: This part of the task that you wrote does the job, that will be done in ISV-914.
Currently we indicate failure/ pass basing on test_results["passed"], and allowed version of preflight/ OCP cluster. 
We should not reimplement this logic in separate task, as in here- you should use some output parameter from step "verify-ci-results".

Allda (approved):

ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
wcheang: It's not verifying results though, just getting them to set the github status. But I'll take a look at what Harita has on the verify-ci-results task to see how this could be simplified.

wcheang: Updated according to the agreed upon change. Please review again. @haripate just FYI I put some placeholder value in verify-ci-results just so that the pipeline is not broken.

ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
```diff
@@ -0,0 +1,70 @@
...
+    - name: github-bot-token
+  steps:
+    - name: set-github-status
+      image: quay.io/redhat-isv/operator-pipelines-images:latest
+      script: |
+        #! /usr/bin/env bash
+        # Don't use set -x to avoid exposing github token
+        set -e
+
+        PREFLIGHT_RESULTS_EXISTS="$(params.preflight_results_exists)"
```
MarcinGinszt, line 37: nitpick: you could use `steps.environment` section like
```
steps:
   - name:  set-github-status
     environment:
       - name: PREFLIGHT_RESULTS_EXISTS
         value: $(params.preflight_results_exists)
```

ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
```diff
@@ -0,0 +1,70 @@
...
+        if [ $PREFLIGHT_RESULTS_EXISTS == "true" ]; then
+          echo "Preflight results already exists - skipping setting github status"
+          exit 0
+        fi
+
+        DESCRIPTION="$(params.description)"
+
+        if [ -z "$(params.test_result)" ]; then
```
MarcinGinszt, line 45: Should we really difference between `test_result` and other states? The only difference I noticed is the difference in the description, and we can enforce that from the pipeline level. 
This way we would also avoid passing arguments such as "result_url", and this way maintain the task more generic

ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
MarcinGinszt: I would also change the `PREFLIGHT_RESULTS_EXISTS` to `SKIP`, to keep the task generic :)

ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
wcheang: Great point, it can be combined with `state` now, and it'll make the task simpler as well. Will go through everything to try to make the task more generic.

ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
wcheang: This is actually not needed anymore, since we can just use the new `params.skip` directly

README.md
```diff
@@ -94,6 +94,7 @@ tkn pipeline start operator-hosted-pipeline \
   --param git_fork_url=https://github.com/MarcinGinszt/operator-pipelines-test.git \
   --param git_repo_url=https://github.com/redhat-openshift-ecosystem/operator-pipelines-test.git \
   --param git_username=test_user \
+  --param git_commit=0aeff5f71e4fc2d4990474780b56d9312554da5a \
```
Allda, line 97: We should be able to get commit hash from already provided params.

MarcinGinszt: No need to, since it's provided by webhook. It would be an unnecessary effort to write and maintain this logic.
However, new parameter should be included in the webhook definition: https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/tasks/operator-hosted-pipeline-trigger.yml#L77

Allda: This will be also needed in the release pipeline and calling a one git command is I guess easier than updating params in few places. I know it is available through webhooks, but as you can see this is an example situation when new param was created and forgotten to add in triggers.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -5,12 +5,15 @@ metadata:
   name: operator-hosted-pipeline
 spec:
   params:
+    # When adding a new param, make sure to also add it to the webhook in
```
wcheang, line 8: There wasn't really a good place in the README, so I decided to add a comment in the pipeline file instead. It's probably more visible this way.

MarcinGinszt (approved):
Great job, thanks!

PR #42, opened 2021-09-03
Task to publish image to ocp registry

MarcinGinszt:


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -126,37 +126,32 @@ spec:
...
           value: "$(tasks.get-cert-project-related-data.results.dist_method)"
-        # params necessary to pull the image
-        - name: registry
-          value: "$(params.registry)"
-        - name: image_namespace
-          value: "$(params.image_namespace)"
-        - name: package_name
-          value: "$(tasks.operator-validation.results.package_name)"
+        - name: image
+          value: "$(params.registry)/$(params.image_namespace)/$(tasks.operator-validation.results.package_name)-index:$(tasks.operator-validation.results.bundle_version)"
```
haripate, line 145: I think it would be good if we also store this value for future use, wdyt?

haripate (changes requested):

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
MarcinGinszt: No need to store it, if we don't need it yet. It would be like creating a variable because it could be needed in the future- it is not much of an effort to do it in the future, and if we'll add it (pointlessly) now, it can waste the time of confused people who will try to understand why it exists now. 
https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -49,25 +50,29 @@
...
+          metadata:
+            name: operator-pipeline-registry-token-{{ suffix }}
+            labels:
+              app: operator-pipeline
+              suffix: "{{ suffix }}"
+              env: "{{ env }}"
+          data:
+            .dockerconfigjson: "{{ lookup('file', operator_pipeline_quay_token, rstrip=False) | b64encode }}"
+    - name: Link secret with pipeline service account (for imagestream operations)
+      ansible.builtin.shell: |
```
Allda, line 74: Generally using a shell in ansible is not recommended because it is not idempotent. Is there a way how define it using k8s module?

ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -49,25 +50,29 @@
...
+        oc secret link pipeline registry-dockerconfig-secret
```
Allda, line 75: How is this SA created? Does it come with a pipeline operator?

ansible/roles/operator-pipeline/tasks/main.yml
MarcinGinszt: I didn't found any :(

Allda: If not, I am also fine with keeping it as is.

ansible/roles/operator-pipeline/tasks/main.yml
MarcinGinszt: yes, 
```
The OpenShift Pipelines Operator adds and configures a Serviceaccount named pipeline that has sufficient permissions to build and push an image. This ServiceAccount is used by pipelineruns.
```
https://openshift.github.io/pipelines-docs/docs/0.10.5/assembly_creating-applications-with-cicd-pipelines.html

ansible/roles/operator-pipeline/templates/openshift/tasks/publish-to-ocp-registry.yml
```diff
@@ -0,0 +1,116 @@
...
+      description: reference to the existing index image
+    - name: is_latest
+      description: If explicitly set to "true", resulting image will be tagged as "latest"
+    - name: bundle_version
+      description: Operator bundle version.
+  results:
+    - name: container_digest
+  steps:
+    - name: imagestream-manipulations
+      image: quay.io/openshift/origin-cli:4.7
```
Allda, line 25: As we discuss this with Brian last week we should use only images that are owned by teams that work on certification workflow and are stored in a trusted registry under our control. Because the openshift folks can accidentally remove the tag and our pipeline start failing.

ansible/roles/operator-pipeline/templates/openshift/tasks/publish-to-ocp-registry.yml
```diff
@@ -0,0 +1,116 @@
...
+      env:
+        - name: PROJECT_NAME
```
Allda, line 27: Is this necessary if vars are already available as params?

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
haripate: Agreeing with you. Thanks for the clarification!

haripate (approved):
LGTM

Allda (approved):

PR #44, opened 2021-09-06
Publish vendor and repository in release pipeline

Allda:
Publish resource publishes a vendor + container repository.

JIRA: ISV-991

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
@@ -114,6 +114,8 @@ spec:
       params:
         - name: cert_project_id
           value: "$(tasks.certification-project-check.results.certification_project_id)"
+        - name: pyxis_url
+          value: "$(tasks.set-env.results.pyxis_url)"
         - name: pyxis_cert_path
           value: $(workspaces.pyxis-ssl-credentials.path)/operator-pipeline.pem
         - name: pyxis_key_path
@@ -130,6 +132,8 @@ spec:
       params:
         - name: org_id
           value: "$(tasks.get-cert-project-related-data.results.org_id)"
+        - name: pyxis_url
+          value: "$(tasks.set-env.results.pyxis_url)"
         - name: pyxis_cert_path
           value: /workspace/pyxis-ssl-credentials/operator-pipeline.pem
         - name: pyxis_key_path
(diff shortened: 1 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -114,6 +114,8 @@ spec:
       params:
         - name: cert_project_id
           value: "$(tasks.certification-project-check.results.certification_project_id)"
+        - name: pyxis_url
```
MarcinGinszt, line 117: Copying the Alex comment:
```
Rather than hard coding the default workspace path, which could change in a future Tekton release, it'd be safer to pass the relative path within the workspace. From the task you can access the workspace path with some variable substitutions. E.g.:

$(workspaces.<workspaceName>.path)
$(workspaces.<workspaceName>.bound) (for optional workspaces)
@Allda may have already created a tracker for changing this in existing tasks. It'd be good to do for the new tasks in this PR though.
```

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -197,6 +201,19 @@ spec:
       params:
         - name: cert_project_id
           value: "$(tasks.certification-project-check.results.certification_project_id)"
+        - name: org_id
+          value: "$(tasks.get-cert-project-related-data.results.org_id)"
+        - name: pyxis_url
+          value: "$(tasks.set-env.results.pyxis_url)"
+        - name: pyxis_cert_path
+          value: /workspace/pyxis-ssl-credentials/operator-pipeline.pem
```
MarcinGinszt, line 209: Copying the Alex comment:
```
Rather than hard coding the default workspace path, which could change in a future Tekton release, it'd be safer to pass the relative path within the workspace. From the task you can access the workspace path with some variable substitutions. E.g.:

$(workspaces.<workspaceName>.path)
$(workspaces.<workspaceName>.bound) (for optional workspaces)
@Allda may have already created a tracker for changing this in existing tasks. It'd be good to do for the new tasks in this PR though.
```

MarcinGinszt (approved):
LGTM, one small comment

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
Allda: Yes, we talked about this with Alex and I am going to file a ticket for the next print to update cert paths in all tasks.

PR #49, opened 2021-09-09
Verifying the user submitting the bundle

MarcinGinszt:
Related PR: https://github.com/redhat-openshift-ecosystem/operator-pipelines-images/pull/24

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/submission-validation.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/submission-validation.yml
@@ -26,10 +26,6 @@ spec:
         #! /usr/bin/env bash
         set -xe
 
-        FILE_WITH_CONTACTS=contacts.txt
-        touch $FILE_WITH_CONTACTS
-        echo "$(params.contacts)" > $FILE_WITH_CONTACTS
-
         # 1. Verify via regex if PR title follows convention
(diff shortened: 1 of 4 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/submission-validation.yml
```diff
@@ -24,15 +26,21 @@ spec:
...
+        # 3. Verify, if the user which created the PR is allowed to
         # submit the operator
-        # 3. Verify that there is no other open PRs for this operator
+        # 4. Verify that there is no other open PRs for this operator
         verify-submission \
           --pr-title "$(params.git_pr_title)" \
           --available-repositories redhat-openshift-ecosystem/operator-pipelines-test \
           --pr-url $(params.git_pr_url) \
+          --git-username $(params.git_username) \
+          --contacts-file $FILE_WITH_CONTACTS \
```
Allda, line 43: argparse supports multiple values for a single argument so you can completely skip the storing and loading contacts to/from file and pass contacts directly as an argument.

https://docs.python.org/3/library/argparse.html#nargs

MarcinGinszt: Good point! Applied that.

Allda (approved):

PR #51, opened 2021-09-09
Fix ansible deployments

MarcinGinszt:
Patching the ServiceAccount with yaml definition instead of using the  openshift client

### DIFF
--- a/ansible/roles/operator-pipeline/tasks/main.yml
+++ b/ansible/roles/operator-pipeline/tasks/main.yml
@@ -71,7 +71,8 @@
           data:
             .dockerconfigjson: "{{ lookup('file', operator_pipeline_quay_token, rstrip=False) | b64encode }}"
     - name: Link secret with pipeline service account (for imagestream operations)
-      community.general.oc: |
+      raw: |
+        echo $KUBECONFIG &&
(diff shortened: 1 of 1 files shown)


ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -71,8 +71,17 @@
...
+        state: present
+        force: yes
+        namespace: "{{ oc_namespace }}"
+        definition:
+          apiVersion: v1
+          kind: ServiceAccount
+          metadata:
+            name: pipeline
+          secrets:
+            - name: registry-dockerconfig-secret
```
Allda, line 84: Doesn't this override previous secrets for the SA?  Since the SA comes from a pipeline operator it can have predefined secrets.

MarcinGinszt: It doesn't, I tested it

Allda (approved):

ansible/roles/operator-pipeline/tasks/main.yml
Allda: ok, then Is fine.

PR #52, opened 2021-09-09
Update preflight params and change results location

komish:
In our testing with preflight in the pipeline, we changed a few things. 

- The loglevel of info is pretty quiet. Preflight may need to make some adjustments as to what goes into which log level, but the `trace` level gives you the most information re: what broke if something breaks.
- The `preflight_namespace` needs to refer to a namespace that exists in the test cluster, not the tekton cluster. I'm thinking it makes more sense to use `default` and change it dynamically as needed.
- The results.json file is being included in the artifacts directory. I updated the path reference that gets passed to the task's outputs, but I'm not sure if this is what you want, given that a the upload-artifacts task seems to target the artifacts directory in its entirety for one of the steps. I'm thinking that having results exist there might be a problem for the upload. Happy to change this however you think is best.

Let me know what I can do to help.

Signed-off-by: Jose R. Gonzalez <josegonzalez89@gmail.com>

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight.yml
@@ -24,10 +24,10 @@ spec:
       default: "default"
     - name: namespace
       description: The namespace for the preflight config
-      default: "$(context.taskRun.namespace)"
+      default: "default"
     - name: log_level
       description: Preflight logging level
-      default: info
+      default: trace
 
   results:
     - name: log_output_file
@@ -96,6 +96,7 @@ spec:
 
         preflight check operator $(params.bundle_image)
 
+        mv $PFLT_ARTIFACTS/results.json results.json
         echo -n $PFLT_LOGFILE | tee $(results.log_output_file.path)
         echo -n results.json | tee $(results.result_output_file.path)
         echo -n $PFLT_ARTIFACTS | tee $(results.artifacts_output_dir.path)


ansible/roles/operator-pipeline/templates/openshift/tasks/preflight.yml
```diff
@@ -97,5 +97,5 @@ spec:
         preflight check operator $(params.bundle_image)
 
         echo -n $PFLT_LOGFILE | tee $(results.log_output_file.path)
-        echo -n results.json | tee $(results.result_output_file.path)
+        echo -n $PFLT_ARTIFACTS/results.json | tee $(results.result_output_file.path)
```
Allda, line 100: Is the path to result configurable? The line below already defines a path to the artifact directory and uploads everything within the dir. I would rather prefer to store results outside an artifacts dir because otherwise we will end up with a duplicated test result in our database.

komish: @Allda 

It's not currently, but Preflight can make it so. I'll work on implementing.

The alternative to that would have been issuing a `mv $PFLT_ARTIFACTS/results.json <wherever-you-want-it>`. I'm thinking this is a bit more of a workaround than what we'd want to implement here.

To that end, would something like a `PFLT_RESULTFILE=/path/to/result.json` be acceptable? In this case, we'll still default to writing the results file to `$PFLT_ARTIFACTS/results.json` if that' unset. 

WDYT?

komish: I went ahead and updated the PR to mv the artifacts into the place the script expects for the time being as that path should work today out of the box.

Allda (approved):

PR #53, opened 2021-09-10
Adding a version of the pipeline that runs serially and passes e2e

tonytcampbell:
When testing the pipeline e2e on an OCP 4.8 cluster deployed with IPI on AWS I hit issues running task in parallel. This PR is to add an option to run all the tasks serially, which worked successfully on this cluster.  This is likely a temporary change until we can resolve.


wcheang:
I'm gonna go ahead and merge this since it's needed early Monday for the Alpha. I've reviewed it, it seems low risk considering that it's a separate pipeline, and Tony has agreed to take on keeping it in sync with the regular ci pipeline. But feel free to still review it after the fact if you want @haripate @MarcinGinszt @Allda

PR #54, opened 2021-09-10
Removing .txt extension from pyxis secret

tonytcampbell:
This is a nit, but we are setting the pyxis api secret with a .txt.  The docs suggest that the .txt shouldn't be there: 

`oc create secret generic pyxis-api-secret --from-literal PYXIS_API_KEY=< API KEY >`

I suggest we drop the extension and follow the pattern in the documentation.


wcheang:
That's fine with me. The doc also differs from the actual code by having `PYXIS_API_KEY` in all caps instead of lower case. Do you mind fixing the doc to be lower case as well?

tonytcampbell:
Yes. I'll update this PR with the doc change too.

tonytcampbell:
Docs updated as well @wcheang per your suggestion.

wcheang (approved):

PR #58, opened 2021-09-13
Set GitHub status when operator is published

Allda:
Github commit status is set when the release pipeline publishes operator
using IIB.

JIRA: ISV-994

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
@@ -10,6 +10,8 @@ spec:
     - name: git_revision
       description: git revision used to check out project
       default: "main"
+    - name: git_commit
+      description: Git repository HEAD commit
     - name: bundle_path
       description: path indicating the location of the certified bundle within the repository
     - name: is_latest
@@ -28,6 +30,7 @@ spec:
     - name: ssh-dir
       optional: true
     - name: pyxis-ssl-credentials
+    - name: github-bot-token
   tasks:
     # Set environment
     - name: set-env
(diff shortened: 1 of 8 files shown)


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -224,3 +227,24 @@ spec:
           value: "$(tasks.get-supported-versions.results.indices)"
         - name: container_digest
           value: "$(tasks.publish-to-ocp-registry.results.container_digest)"
+
+    - name: set-github-status-published
```
MarcinGinszt, line 231: Well, this task should forward any results of bundle publishing (either successful or not), retrieved from IIB after the publishing call-  Jira says, that it's blocked by ISV-993, and it's done before it.

Allda: I added a extra results to the publishing step and reuse it as a status argument. I am going to work on the publishing step once we have a Kerberos account.

wcheang: When you get to the publishing step, please make sure that the status result of the step are within the list of accepted states of the github status API: https://docs.github.com/en/rest/reference/repos#create-a-commit-status

wcheang (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
Allda: yy, I am aware of it.

PR #61, opened 2021-09-14
Necessary permissions for pipeline Service Account

MarcinGinszt:
Service Account `pipeline` used by openshift pipelines needs permissions for project creation. They are used in task imagestream-manipulation.

### DIFF
--- a/ansible/init.sh
+++ b/ansible/init.sh
@@ -60,16 +60,12 @@ main() {
     for env in $environments;
     do
         execute_playbook $env $passwd_file
-    done
 
-    # Grant the pipeline service account permissions to create projects
-    # -it cannot be done in Playbook, as ansible SA has no permissions to create
-    # cluster-scoped resources- and this command creates ClusterRoleBinding
-    for env in $environments;
-    do
+        # Grant the pipeline service account permissions to create projects
+        # -it cannot be done in Playbook, as ansible SA has no permissions to create
+        # cluster-scoped resources- and this command creates ClusterRoleBinding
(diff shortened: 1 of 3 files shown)


ansible/init.sh
```diff
@@ -62,6 +62,15 @@ main() {
         execute_playbook $env $passwd_file
     done
 
+    # Grant the pipeline service account permissions to create projects
+    # -it cannot be done in Playbook, as ansible SA has no permissions to create
+    # cluster-scoped resources- and this command creates ClusterRoleBinding
+    for env in $environments;
+    do
+        oc adm policy add-cluster-role-to-user self-provisioner -z pipeline -n operator-pipeline-$env
```
Allda, line 70: Can you add the oc command the the loop above? I don't see any benefit of duplicating a code.

ansible/init.sh
```diff
@@ -62,6 +62,15 @@ main() {
         execute_playbook $env $passwd_file
     done
 
+    # Grant the pipeline service account permissions to create projects
```
Allda, line 65: This step requires a pipelines operator to be installed on a cluster. Can you please mention it somewhere in a doc? Or maybe even better install the pipeline operator in an ansible role.

MarcinGinszt: Most of the steps (eg. installing tasks and pipelines) already required the pipeline operator. Can I create an additional ticket to do that?

MarcinGinszt: https://issues.redhat.com/browse/ISV-1115

Allda (approved):

PR #67, opened 2021-09-20
[EET-1918,EET-2143] add preflight-trigger task

komish:
This is an initial task that utilizes preflight-trigger to start a preflight execution in DPTP. This won't work out of the box because it requires specific credentials, etc., so it has not been added to any of the pipelines in this PR. Please also see the mentioned Jira EET-1918 for some decisions we'll need to make before we start using this.

Signed-off-by: Jose R. Gonzalez <josegonzalez89@gmail.com>

cc @mrhillsman

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
@@ -36,6 +36,7 @@ spec:
     - name: github-bot-token
     - name: kubeconfig
     - name: hydra-credentials
+    - name: preflight-decryption-key
   tasks:
     # Set environment
     - name: set-env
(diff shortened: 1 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yaml
```diff
@@ -0,0 +1,64 @@
...
+        - '-pflt-index-image'
+        - $(params.bundle_index_image)
+        - '-test-asset'
+        - $(params.bundle_image)
+        - '-asset-type'
+        - 'operator'
+        - '-output-path'
+        - 'preflight-output.json'
+      env:
+        - name: KUBECONFIG
```
MarcinGinszt, line 54: Why do we need Kubeconfig? How can we supply it- to which cluster will it point?
We expected this task to output the Kubeconfig so that we could connect to some cluster and run the preflight there.

komish: This is the kubeconfig preflight-trigger uses to apply the prowjob to the DPTP cluster - we'll discuss how to get that over to you.

jpaulovic:
@komish @MarcinGinszt @amisstea I haven't seen any activity on this PR since 8 days ago. Is there anything remaining to do before this can be merged? Also, are you going to add the necessary parameters to the pipeline config or are you expecting us to do that? (both is fine, we're just blocked by this right now)

komish:
/do-not-merge hold

@jpaulovic This is just waiting for the work in #79 to finish. I don't think either one contains the pipeline config, but we'll add that as well.

komish:
@jpaulovic @amisstea This is ready for review, and includes the pipeline changes.

This PR will:

- Add a preflight-trigger task, which runs Preflight in DPTP. This then pulls encrypted results from the DPTP Execution, decrypts them, and places them in the associated "results" locations for the upload task. 
- Rewires the Preflight task in the hosted pipeline to use preflight-trigger as the taskRef instead of preflight.

haripate (approved):
LGTM

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -441,28 +442,29 @@ spec:
...
 
-    # TODO: passing kubeconfig here is just a temporary workaround to make the task pass
     - name: preflight
       runAfter:
         - ocp-environment-preparation
       taskRef:
-        name: preflight
+        kind: Task
+        name: preflight-trigger
       params:
```
kosciCZ, line 451: AFAIK the preflight should run for the latest supported version of the operator:

```
        - name: ocp_version
          value: "$(tasks.get-supported-versions.results.max_supported_ocp_version)"
```

ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
```diff
@@ -0,0 +1,116 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: preflight-trigger
+spec:
+  params:
+    - default: '4.8'
```
kosciCZ, line 7: I'd probably avoid setting a default here, I think we'll always supply it (see my other comment)

kosciCZ (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
mrhillsman: Yes we are not going to set the at all and are going to depend on the max_supported_ocp_version as you noted in your previous comment. It is hardcoded for now to ensure we run against a version dptp definitely has in their Hive clusterpool.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
mrhillsman: See https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/67#discussion_r719554484

komish:
@mrhillsman per request - the task name has changed to be preflight-trigger, and result references have been updated to match.

ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
komish: Indeed. I think we need to land on a list of supported versions _today_, after which @mrhillsman and I can make sure DPTP has the right configurables that preflight-trigger can reach into before we remove this default value.

kosciCZ (approved):

PR #74, opened 2021-09-23
[ISV-1090] Digest pinning should name the generated branch after the original ref

haripate:


ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -78,8 +78,8 @@ spec:
         if [ $DIRTY_FLAG == "true" ]; then
           echo "Commit changes to temporary branch"
 
-          RANDOM_STRING=$(cat /dev/urandom | tr -dc 'a-z0-9'| fold -w 3 | head -n 1)
-          TEMP_BRANCH="pinned-operator-manifest-${RANDOM_STRING}"
+          CURRENT_BRANCH=$(git rev-parse --abbrev-ref HEAD)
+          TEMP_BRANCH="${CURRENT_BRANCH}-pinned"
```
amisstea, line 82: Because this is no longer random, we should add `-f` to the `git push` call below. This will allow users to re-run the pipeline without needing to cleanup the pinned branch should this task be repeated.

amisstea (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
haripate: Updated!

MarcinGinszt: If this branch will already exist, then `git checkout -b` will fail the script.
Shouldn't we firstly do `git branch -D "${TEMP_BRANCH}" || true`?

amisstea: It may exist on the remote repository but there shouldn't be a local branch since the workspace is recreated for each `PipelineRun`.

amisstea (approved):

amisstea:
LGTM :+1:

PR #59, opened 2021-09-13
Create Container Image

MarcinGinszt:
Related PR: https://github.com/redhat-openshift-ecosystem/operator-pipelines-images/pull/26
Jira task: IVS-1036

PR #75, opened 2021-09-23
Add step to call IBM webhook to trigger marketplace replication

wcheang:


wcheang:
This will probably conflict with https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/59 due to the &bundleImage

ansible/inventory/group_vars/operator-pipeline.yml
```diff
@@ -16,6 +16,7 @@ operator_pipeline_registry_token: ../../vaults/{{ env }}/dockerconfig.json
 
 operator_pipeline_kubeconfig_prow_local_path: ../../vaults/common/kubeconfig-prow
 operator_pipeline_github_bot_token: ../../vaults/common/github-bot-token.txt
+operator_pipeline_ibm_webhook_token_local_path: ../../vaults/common/nonprod-ibm-webhook-token
```
amisstea, line 19: The prod and preprod token is the same? :weary:

ansible/roles/operator-pipeline/templates/openshift/tasks/trigger-marketplace-replication.yml
```diff
@@ -0,0 +1,42 @@
...
+      workingDir: $(workspaces.source.path)
+      env:
+        - name: WEBHOOK_TOKEN_PATH
+          value: $(workspaces.ibm-webhook-token.path)/token
+      script: |
+        #! /usr/bin/env bash
+        # Don't use set -x to avoid exposing webhook token
+        set -e
+
+        export IBM_WEBHOOK_TOKEN=$(cat $WEBHOOK_TOKEN_PATH)
```
amisstea, line 34: Is it possible to use `secretKeyRef` instead of doing this `export`? Take a look at [this](https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml#L35-L39) for an example. There is a very real risk of leaking credentials this way so we should try to avoid it if possible.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -286,6 +292,35 @@ spec:
         - name: ocp-registry-kubeconfig
           workspace: ocp-registry-kubeconfig
 
+    # Trigger IBM webhook for marketplace replication
+    - name: trigger-marketplace-replication
+      runAfter:
+        - publish-to-ocp-registry
```
haripate, line 298: I think we also require the task `trigger-marketplace-replication` should run after the `operator-validation` task.

ansible/roles/operator-pipeline/templates/openshift/tasks/trigger-marketplace-replication.yml
```diff
@@ -0,0 +1,46 @@
...
+      workingDir: $(workspaces.source.path)
+      env:
+        - name: IBM_WEBHOOK_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: $(params.ibm_webhook_token_secret_name)
+              key: $(params.ibm_webhook_token_secret_key)
+      script: |
+        #! /usr/bin/env bash
+        # Don't use set -x to avoid exposing webhook token
```
amisstea, line 37: No need to keep this comment now :smile:

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -348,6 +383,7 @@ spec:
       runAfter:
         - publish-resources
         - get-supported-versions
+        - trigger-marketplace-replication
```
amisstea, line 387: I think you can remove `get-supported-version` since `trigger-marketplace-replication` depends on that. The `runAfter` clauses dictate what the pipeline diagram looks like so it'd be nice to fix this.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
wcheang: Yeah it should, since it depends on request from `operator-validation`. Good catch! Added now.

amisstea: Seems like it. While this would work it's nice to see the explicit dependencies in the diagram.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
wcheang: Removed. The pipeline graph looks simpler now :laughing:

Allda (approved):

PR #78, opened 2021-09-28
Use the OCP clusters to publish the image

MarcinGinszt:
Jira task: ISV-1134


ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -50,38 +50,24 @@
...
+  k8s:
+    state: present
+    force: yes
+    namespace: "{{ oc_namespace }}"
+    definition:
+      apiVersion: v1
+      kind: Secret
+      type: kubernetes.io/dockerconfigjson
+      metadata:
+        name: registry-dockerconfig-secret
```
amisstea, line 64: Is this supposed to be the same secret as the one from the instructions in `docs/rhc4tp-cluster.md`? I ask because that's not a secret for quay but rather the registry cluster. Do we need to use different names to avoid confusion or is there more to this?

MarcinGinszt: I feel I'm missing smething- that's a Quay token. It's used to push the bundle and index images to quay, and later to retrieve them from quay in rhc4tp cluster.

amisstea (approved):
LGTM :+1:

PR #79, opened 2021-09-28
Add preflight artifact decryption task

mrhillsman:
Task that expects preflight-job-url returned from preflight task on
successful run, appends path to encrypted artifacts from preflight,
obtains and decrypts the artifacts using GPG.


mrhillsman:
moved to #67

PR #80, opened 2021-09-29
Webhook to trigger the release pipeline

MarcinGinszt:


ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -183,4 +183,6 @@
                     command: ["tkn", "pipelinerun", "delete", "--keep", "{{ preserve_after_cleanup }}"]
                 restartPolicy: OnFailure
 
-- include: tasks/operator-hosted-pipeline-trigger.yml
+- include: tasks/webhook-event-listener.yml
+- include: tasks/webhook-hosted-pipeline-trigger.yml
+- include: tasks/webhook-release-pipeline-trigger.yml
```
amisstea, line 188: We can do this as a followup but should we move all these into a more appropriately named directory since they're not tasks? Perhaps `triggers`.

ansible/roles/operator-pipeline/tasks/webhook-event-listener.yml
```diff
@@ -0,0 +1,107 @@
...
+                        (
+                          header.match("X-GitHub-Event", "pull_request")
+                          && body.action in ["opened", "reopened", "synchronize"]
+                          && body.pull_request.base.ref == "{{ branch }}"
+                        )
+                bindings:
+                  - ref: operator-hosted-pipeline-trigger-binding
+                template:
+                  ref: operator-hosted-pipeline-trigger-template
+              # run Hosted pipeline on PR labeled with label "pipeline/trigger"
```
amisstea, line 42: Do we need a similar listener for the release pipeline? I'm assuming this is for manual intervention in the case of a failure so it may be useful to have the same for the release pipeline because we won't be able to change the PR state.

ansible/roles/operator-pipeline/tasks/main.yml
MarcinGinszt: The name `tasks` exists here in the context of Ansible, as an "ansible thing to do"
https://docs.ansible.com/ansible/latest/user_guide/playbooks_intro.html#task-execution
We have this convention in other projects too
https://gitlab.cee.redhat.com/isv/mercury/mercury/-/tree/master/ansible/roles/mercury/tasks

ansible/roles/operator-pipeline/tasks/webhook-event-listener.yml
MarcinGinszt: Done, there are now listeners for labels pipeline/trigger-hosted and pipeline/trigger-release

ansible/roles/operator-pipeline/tasks/webhook-release-pipeline-trigger.yml
```diff
@@ -0,0 +1,96 @@
...
+                    - name: git_pr_title
+                      value: $(tt.params.git_pr_title)
+                    - name: is_latest
+                      value: "true"
+                  workspaces:
+                    - name: repository
+                      volumeClaimTemplate:
+                        spec:
+                          accessModes:
+                            - ReadWriteOnce
```
Allda, line 70: Doesn't this make parallel tasks impossible?

ansible/roles/operator-pipeline/tasks/webhook-event-listener.yml
```diff
@@ -0,0 +1,127 @@
+---
```
Allda, line 1: I think this and other files from this PR should not be stored in `tasks` directory because it is not a tekton task. Can you make separate dir for triggers?

ansible/roles/operator-pipeline/tasks/webhook-release-pipeline-trigger.yml
MarcinGinszt: Changed all occurences of ReadWriteOnce to ReadWriteMany

ansible/roles/operator-pipeline/tasks/webhook-event-listener.yml
Allda: Aha, i didn't noticed it is a separate task directory.

Allda (approved):

PR #84, opened 2021-10-04
[ISV-1198] Fix up the result value for get cert project related data Task

haripate:


ansible/roles/operator-pipeline/templates/openshift/tasks/get-cert-project-related-data.yml
```diff
@@ -57,3 +60,5 @@ spec:
         cat cert_project | jq '.org_id' | tr -d '"' |  tr -d '"' | grep . | tr -d "\n" | tee $(results.org_id.path)
         cat cert_project | jq '.contacts[].email_address' | tr '\n' ' ' | tr -d '"' | grep . | tr -d "\n" | tee $(results.contacts.path)
         cat cert_project | jq '.operator_distribution' |  tr -d '"' | grep . | tr -d "\n" | tee $(results.operator_distribution.path)
+        echo -n "none" | tee $(results.certification_status.path)
+        echo -n "false" | tee $(results.update_cert_project_status.path)
```
amisstea, line 64: If the current certification_status is `Waiting for container image` (or a number of other possible allowed values), this won't correct it which isn't ideal. I think the logic may need to be changed to the following:

```
#pseudo code
if current_status != "Published":
    new_status = "In Progress"
    update_status = true
else:
   new_status = none
   update_status = false
```

ansible/roles/operator-pipeline/templates/openshift/tasks/get-cert-project-related-data.yml
```diff
@@ -15,6 +15,8 @@ spec:
     - name: pyxis_url
       default: https://pyxis.engineering.redhat.com
   results:
+    - name: update_cert_project_status
+      description: The boolean value indicating if the cert project status needs to be updated or not
```
amisstea, line 19: What about changing the result name/description for `certification_status`? It's misleading right now because I'd expect that to return the current status rather than a _possible_ future status. It would be useful to return both current and future states to clear up any potential confusion.

ansible/roles/operator-pipeline/templates/openshift/tasks/update-cert-project-status.yml
```diff
@@ -33,13 +35,17 @@ spec:
...
-          --cert-project-id $CERT_PROJECT_ID \
-          --certification-status "$CERTIFICATION_STATUS"
+        if [ "$UPDATE_CERT_PROJECT_STATUS" == "true" ]; then
+          update-cert-project-status \
+            --pyxis-url $PYXIS_URL \
+            --cert-project-id $CERT_PROJECT_ID \
+            --certification-status "$CERTIFICATION_STATUS"
 
-        echo "Updating the Certification Project status to $CERTIFICATION_STATUS"
+          echo "Updating the Certification Project status to $CERTIFICATION_STATUS"
```
amisstea, line 50: There should be a human readable log message for the else condition as well.

amisstea (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/get-cert-project-related-data.yml
haripate: I was following this AC info https://issues.redhat.com/browse/ISV-871. In this task, it is mentioned that only if the state is `Started` then we need to update the state to `In Progress`.

amisstea: Unfortunately that's no longer sufficient because of the prevalence of other incorrect/invalid states in the data.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-cert-project-related-data.yml
```diff
@@ -48,6 +50,15 @@ spec:
           --pyxis-url $(params.pyxis_url) \
           --cert-project-id $(params.cert_project_id)
 
+        CURRENT_CERTIFICATION_STATUS=$(cat cert_project | jq '.certification_status'| tr -d '"')
```
amisstea, line 53: You can probably remove the `| tr -d '"'` portion of this command if you use `jq -r`.

amisstea (approved):
Only one minor suggestion. Otherwise LGTM :+1:

PR #89, opened 2021-10-06
Renew github token

Allda:
The previous token was already expired. The new one doesn't expires.

JIRA: ISV-1192


MarcinGinszt (approved):
LGTM! Thanks

PR #93, opened 2021-10-08
various bugfixes

MarcinGinszt:
Various bugfixes:
- min versions set to 0.0.0 (untill we know, what should we expect)
- passing parameter `env` to release pipeline through trigger
- removing unused task `ocp-environment-preparation`
- get-ci-results uses quoted envs, which resolves issue with multiline env
- adding missing results to publish-bundle
- mention a few other bugs


amisstea (approved):
LGTM :+1:

PR #100, opened 2021-10-12
[ISV-1157] Pin digest for operator-pipelines-images

haripate:


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline-serial.yml
```diff
@@ -38,6 +38,9 @@ spec:
     - name: submit
       description: Set to "true" to submit results and open a pull request.
       default: "false"
+    - name: image_tag
```
amisstea, line 41: Can we rename this, and the task params, to `pipeline_image` instead? It's not just a tag nor does it need to contain one. It can be any valid pull spec.

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: certification-project-check
 spec:
   params:
+    - name: image_tag
```
amisstea, line 8: This param isn't being used in all these tasks. Do you intend to fix that?

amisstea (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
haripate: sure

amisstea (approved):
LGTM :+1:

PR #106, opened 2021-10-13
Fix how results are passed around

amisstea:
This also fails the hosted pipeline intentionally when the preflight results don't pass.


haripate (approved):
LGTM

wcheang (approved):

PR #108, opened 2021-10-13
Set pending GitHub status at the start of the pipelines

amisstea:


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -45,8 +45,32 @@ spec:
     - name: hydra-credentials
     - name: preflight-decryption-key
   tasks:
+    # Set a pending GitHub status check first to provide the user with immeditate feedback.
+    - name: set-github-status-pending
+      taskRef:
+        name: set-github-status
+      params:
+        - name: pipeline_image
```
haripate, line 53: I think the `target_url` param is missing here.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -47,8 +47,32 @@ spec:
     - name: ocp-registry-kubeconfig
 
   tasks:
+    # Set a pending GitHub status check first to provide the user with immeditate feedback.
+    - name: set-github-status-pending
+      taskRef:
+        name: set-github-status
+      params:
+        - name: pipeline_image
```
haripate, line 55: Same here.

PR #110, opened 2021-10-13
[ISV-1219] The github status checks should include a "details" link to the connect portal interface for the preflight results and pipeline logs

haripate:


ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-status.yml
```diff
@@ -54,4 +54,5 @@ spec:
           --commit-sha "$(params.commit_sha)" \
           --status "$(params.state)" \
           --context "$(params.context)" \
-          --description "$(params.description)"
+          --description "$(params.description)" \
+          --target-url "$(params.target_url)"
```
amisstea, line 58: In relation this [this comment](https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/108/files#r728431971), what will happen if the default empty string is provided here? The desired end result is no "Details" link appears on the status check.

haripate: I just pushed new changes with it. I put one condition to check whether the target_url param value is empty or not.

wcheang: Just did a quick test with curl, it looks like the Github API will work fine with empty string as the target-url value. It'll just not show the "Details" link.

haripate: The tekton task `set-github-status` fails if we try to pass an empty string as a parameter to the function. Here is the Example of failed task: https://console-openshift-console.apps.pipelines-stage.0ce8.p1.openshiftapps.com/k8s/ns/haripate-playground/tekton.dev~v1beta1~PipelineRun/operator-hosted-pipeline-run-hjz7d/logs/set-github-status-failure

wcheang: I don't think that has anything to do with the empty string. The error says `set-github-status: error: unrecognized arguments: --target-url `, and target-url is also not showing up in the usage output. The operator-pipeline-images change was probably just not pushed to the latest container image yet.

haripate: Oh, I got it now. Updated.

amisstea (approved):

PR #118, opened 2021-10-18
Make quay bundle and index repos public in hosted pipeline

amisstea:


ansible/roles/operator-pipeline/tasks/main.yml
```diff
@@ -153,8 +153,8 @@
           suffix: "{{ suffix }}"
           env: "{{ env }}"
       data:
-        username: "{{ quay_username | b64encode }}"
-        password: "{{ quay_password | b64encode }}"
+        username: "{{ iib_quay_username | b64encode }}"
+        password: "{{ iib_quay_password | b64encode }}"
```
amisstea, line 157: These were renamed across all the `secret-vars.yml` files to avoid confusion with `quay_oauth_token`.

PR #121, opened 2021-10-20
Remove dependency variables from set-status

Allda:
When pipeline fails we always want to set github failed status. This
removal allows to set status always no matter when a pipeline fails.

JIRA: ISV-1269


Allda:
@amisstea Let me know if this is a good solution. The result dependency doesn't allow to trigger a task in finally section so the removal solves the issue.

PR #126, opened 2021-10-21
Add task to check that cluster version is supported

wcheang:


ansible/roles/operator-pipeline/templates/openshift/tasks/check-ocp-cluster-version.yml
```diff
@@ -0,0 +1,35 @@
...
+      script: |
+        #! /usr/bin/env bash
+        set -xe
+
+        KUBECONFIG=$(workspaces.kubeconfig.path)/kubeconfig
+        CLUSTER_VERSION=$(oc get clusterversion -o jsonpath="{.items[0].spec.desiredUpdate.version}" --kubeconfig $KUBECONFIG)
+        # Grab only the major and minor version for comparison
+        CLUSTER_VERSION=$(echo $CLUSTER_VERSION | cut -d. -f1,2)
+        # parse indices into a list of strings with OCP versions
+        SUPPORTED_VERSIONS=$(echo '$(params.indices_ocp_versions)' | jq '.[]' | tr -d '"')
```
Allda, line 25: you can use `jq -r` to return a raw string and then you don't need to use `tr`

PR #146, opened 2021-10-28
Script to set- up the test environment for the developers

MarcinGinszt:
Jira task: ISV-1209


ansible/inventory/group_vars/operator-pipeline.yml
```diff
@@ -1,6 +1,6 @@
 ##### General settings
 env: unset
-oc_namespace: operator-pipeline-{{ env }}
+oc_namespace: "{{ (namespace) if (namespace) is defined else operator-pipeline-(env) }}"
```
Allda, line 3: If you pass the oc_namespace/namespace as env variable when running a playbook the value from inventory gets overwritten and you don't have to do this.

ansible/init-custom-env.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
+
+# Creates service account for each operator-pipeline env; updates secret-vars.yml with token
+#
+# As arguments, it expects environments for which the sa should be created
+#
+# ./init.sh dev stage prod
+#       creates the sa for dev, stage and prod environments
```
MarcinGinszt, line 8: Just noticed the deprecated comment, updating it now

Allda (approved):
One minor detail, but otherwise it looks ok

ansible/inventory/group_vars/operator-pipeline.yml
MarcinGinszt: Good point! Applied.

PR #157, opened 2021-11-02
[ISV-933] Merge the operator-pipeline-images code to operator-pipeline code

haripate:
**The Repo structure is shown below:**

Screenshot from 2021-11-10 17-31-23


.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
+---
```
MarcinGinszt, line 1: Could you change the name of this file to something more descriptive? Now it's not a main workflow, it's just the image workflow :)

.gitignore
```diff
@@ -13,3 +13,146 @@ kubeconfig
...
+
+# mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
+
+# Pyre type checker
+.pyre/
+
+# VSCode project settings
```
MarcinGinszt, line 150: Stuff below is duplicated, it already existed in the .gitignore

.github/workflows/main.yml
```diff
@@ -0,0 +1,48 @@
...
+name: Build and Push Image
+
+on:  # yamllint disable-line rule:truthy
+  - push
```
MarcinGinszt, line 5: @Allda I know it was working this way in the `operator-pipelines-images`, but should we really build and push image on each push to repo?

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
+---
```
MarcinGinszt, line 1: Again, could you give it more descriptive name? Eg. release-image.yml

.dockerignore
```diff
@@ -0,0 +1,8 @@
+# Ignoring git and cache folders
```
MarcinGinszt, line 1: We don't need the whole pipelines code in the image- you can also ignore `templates` and `ansible`.

MarcinGinszt (changes requested):
I think that all the files that are necessary only for the `operator-pipeline-images`, should be in the directory `operator-pipeline-images` I mean `certs`, `requirements.txt`, `requirements-dev.txt` and- if you agree with me- also setup.py and dockerfile. WDYT?

.github/workflows/main.yml
Allda: Yeah, why not? This allows users to build images and test them without manually building and pushing images to the registry.

wcheang:
Files like requirements*.txt and tox.ini should be moved to the root level and adjusted accordingly

amisstea (approved):

wcheang (approved):

wcheang:
If the resulting image has been tested to work, then LGTM :+1:

PR #158, opened 2021-11-05
E2E test of CI pipeline

MarcinGinszt:
Jira task: ISV-1108
Now CI pipeline runs in parallel with deployment on merge to main. Can also be triggered manually.
To full E2E run, we are missing the Pyxis certificate- Jira task ISV-1360


.github/workflows/e2e.yml
```diff
@@ -0,0 +1,66 @@
...
+          # Prepare Kubeconfig.
+          # Use GitHub secret which contains the Kubeconfig for SA with admin privilleges (same as in ansible vault)
+          # to create Kubernetes context.
+          mkdir $HOME/.kube
+          cat <<EOF > $HOME/.kube/config
+          ${{ secrets.KUBECONFIG }}
+          EOF
+
+          # Prepare vault password
+          echo ${{ secrets.VAULT_PASSWORD }} > ansible/vault-password
```
amisstea, line 37: This seems insecure to me on the off chance these commands end up in logs. What can be done to ensure these won't appear in GitHub at all?

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          echo "OC_PROJECT=test-e2e-$GITHUB_SHA" >> $GITHUB_ENV
+      - uses: actions/checkout@v1
+      - name: Prepare environment
+        run: |
+          # Install dependencies
+          pip install --user openshift
+
+          # Install tkn cli
+          curl -LO https://github.com/tektoncd/cli/releases/download/v0.11.0/tkn_0.11.0_Linux_x86_64.tar.gz
```
amisstea, line 25: Why use such an old version? 0.17.2 is what I have from OCP 4.7/4.8.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,66 @@
...
+            --param env=stage \
+            --workspace name=pipeline,volumeClaimTemplateFile=templates/workspace-template.yml \
+            --workspace name=kubeconfig,secret=kubeconfig \
+            --workspace name=ssh-dir,secret=github-ssh-credentials \
+            --workspace name=pyxis-api-key,secret=pyxis-api-secret \
+            --showlog
+      - name: Check if pipeline passed
+        run: |
+          # get second line of second column from output- second column is SUCCEEDED, second line- value
+          SUCCEEDED=$(oc get pr | awk 'FNR==2{print $2}')
```
amisstea, line 59: How about using `-o json` with a little `jq` instead?

.github/workflows/e2e.yml
MarcinGinszt: Correct me if I'm wrong, but I'm not seeing any vulnerability here. If it would appear in the logs, it should be redacted by GitHub. 
GitHub security best practices are not opposing this approach.
https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets
They, however, are mentioning, that secret should not be structured data, like our Kubeconfig secret is, as there is a risk that the data won't be detected in the terminal output. However, I tested that by adding secret TEST_SECRET with mock Kubeconfig. It was redacted just fine
https://github.com/redhat-openshift-ecosystem/operator-pipelines/runs/4136073792?check_suite_focus=true#step:4:126

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,71 @@
+---
+name: E2E-CI
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - main
+      - ISV-1108-2
```
Allda, line 8: Don't forget to remove it before merging the PR.

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,71 @@
+---
+name: E2E-CI
+
+on:  # yamllint disable-line rule:truthy
+  push:
```
Allda, line 5: If you want this job to run as a CI of each PR you should also add `pull_request` here.

Allda (approved):
Few minor details, but it looks good to me.

PR #160, opened 2021-11-08
Reduce the chances of the hosted pipeline not uploading logs

amisstea:
The operator-validation task was renamed to bundle-path-validation to
better reflect what it verifies. When uploading logs from the hosted and
release pipelines, the package name and bundle version are taken from
verify-pr-title because it's the earliest moment these values are known.
Hence increasing the odds that logs will be available to the partner.


Allda (approved):

MarcinGinszt (approved):
LGTM

PR #164, opened 2021-11-10
[ISV-1344] As a partner, the hosted pipeline won't re-test my operator if only project settings are changed

haripate:


ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results-attempt.yml
```diff
@@ -71,7 +71,8 @@ spec:
 
         if [ -f $RESULTS_FILE ]; then
           echo "Results exist"
-          echo -n "true" | tee $(results.preflight_results_exists.path)
+          CHECK_RESULTS=$(cat $RESULTS_FILE | jq -r '.passed')
```
amisstea, line 74: Please add a log message to tell the user if these results passed.

haripate: Updated!

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results-attempt.yml
```diff
@@ -72,6 +72,10 @@ spec:
         if [ -f $RESULTS_FILE ]; then
           echo "Results exist"
           CHECK_RESULTS=$(cat $RESULTS_FILE | jq -r '.passed')
+          if [ $CHECK_RESULTS = "false" ]; then
+            echo "Not all preflight tests passed."
+          fi
+          echo "All preflight tests passed successfully."
```
amisstea, line 78: Shouldn't this be inside an else condition?

haripate: Doesn't this work as else condition?

amisstea (approved):

PR #166, opened 2021-11-11
Adding an explicit check for relatedImages

tonytcampbell:
With the resent modification to allow partners to manually pin their digest, we were not checking for relatedImages in that case.  RelatedImages are required for Marketplace.  This check uses the digest pinning tool to get the number of images in the CSV and make sure there is an equal number of images listed in the relatedImages.  In a future PR we can use a more robust check that does more than comparing image counts in the CSV to image counts in the relatedImages section.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
@@ -11,6 +11,7 @@ spec:
     - name: pipeline_image
   results:
     - name: dirty_flag
+    - name: related_images_flag
   workspaces:
     - name: source
     - name: registry-credentials
@@ -78,4 +79,30 @@ spec:
         else
           echo "Manifests are pinned."
           echo -n "false" | tee $(results.dirty_flag.path)
+        fi
+    - name: verify-related-images
+      image: "$(params.pipeline_image)"
+      workingDir: $(workspaces.source.path)
+      script: |
+        #! /usr/bin/env bash
+        set -xe
+
+        if [ "$(params.enabled)" != "true" ]; then
+          echo "Digest pinning is not enabled"
+          echo -n "true" | tee $(results.related_images_flag.path)
+          exit 0
+        fi
+
+        BUNDLE_PATH=$(realpath $(params.bundle_path))
+        cat references.json
+        REFERENCE_COUNT=$(jq length references.json)
+        CSVFILE=$(find $BUNDLE_PATH -name "*clusterserviceversion.yaml" -o -name "*clusterserviceversion.yml")
+        RELATED_IMAGE_COUNT=$(yq -e '.spec.relatedImages | length' $CSVFILE)
+
+        if [[ $REPLACEMENT_COUNT == $RELATED_IMAGE_COUNT ]]; then
+          echo "Related images section exists."
+          echo -n "true" | tee $(results.related_images_flag.path)
+        else
+          echo "Related images section does not exist."
+          echo -n "false" | tee $(results.related_images_flag.path)
         fi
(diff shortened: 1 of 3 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
```diff
@@ -79,3 +80,23 @@ spec:
...
+        set -xe
+
+        BUNDLE_PATH=$(realpath $(params.bundle_path))
+        cat references.json
+        REFERENCE_COUNT=$(jq length references.json)
+        CSVFILE=$(find $BUNDLE_PATH -name "*clusterserviceversion.yaml" -o -name "*clusterserviceversion.yml")
+        RELATED_IMAGE_COUNT=$(yq -e '.spec.relatedImages | length' $CSVFILE)
+
+        if [[ $REPLACEMENT_COUNT == $RELATED_IMAGE_COUNT ]]; then
+          echo "Related images section exist."
```
amisstea, line 97: Minor grammar change: "exist" -> "exists"

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
```diff
@@ -79,3 +80,23 @@ spec:
           echo "Manifests are pinned."
           echo -n "false" | tee $(results.dirty_flag.path)
         fi
+    - name: verify-related-images
+      image: "$(params.pipeline_image)"
+      workingDir: $(workspaces.source.path)
+      script: |
```
amisstea, line 86: Shouldn't this include a check for `params.enabled` like the previous steps have?

amisstea (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
tonytcampbell: Good question.  Now that I look at it, I think we only need to check `params.enabled` in the `pin-digest` step.  The `verify-pinning` and the `verify-related-images` steps should run regardless of `params.enabled`.  `params.enabled` == false doesn't necessarily mean `dirty_flag` == true anymore since we support manual pinning as well.  I'll update this to remove the `params.enabled` from the `verify-pinning` step.

amisstea: But will that work for a CI pipeline run where pinning is disabled? We've found that it's necessary to propogate skipping behavior across all steps of a task if other tasks will depend on it from a pipeline.

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
tonytcampbell: Thanks for catching that. I changed it.

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
tonytcampbell: You're right.  I added the `params.enabled` checks for `verify-pinning` and `verify-related-images`.  Thanks for the review.

wcheang:
Just have one question about the cat statement, but otherwise looks good :+1:

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
```diff
@@ -79,3 +80,29 @@ spec:
...
+        set -xe
+
+        if [ "$(params.enabled)" != "true" ]; then
+          echo "Digest pinning is not enabled"
+          echo -n "true" | tee $(results.related_images_flag.path)
+          exit 0
+        fi
+
+        BUNDLE_PATH=$(realpath $(params.bundle_path))
+        cat references.json
```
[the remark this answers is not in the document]
tonytcampbell: Just debugging.  Let me know if you think it's too noisy.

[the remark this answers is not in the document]
amisstea: We can evaluate logging clarity/noise in a separate issue, if necessary. I'm going to resolve this in favor of getting this hotfix out.

amisstea (approved):

PR #168, opened 2021-11-11
Allowing the ci.yaml to be modified

tonytcampbell:
Currenty, STEP-VERIFY-CHANGED-DIRECTORIES will fail if a PR includes changes to the `ci.yaml` file. The current implementation rejects any files in a PR that do not have a status of `added` therefore all `modified` files are rejected. This patch adds a single exception to that rule for the `ci.yaml` file.  This will allow users to set `merge:false` in `ci.yaml` and covers a few edge cases where a `cert_projet_id` may need to be modified.

### DIFF
--- a/operator-pipeline-images/operatorcert/__init__.py
+++ b/operator-pipeline-images/operatorcert/__init__.py
@@ -218,24 +218,30 @@ def get_files_added_in_pr(
     rsp = requests.get(compare_changes_url)
     rsp.raise_for_status()
 
-    added_files_names = []
+    added_files = []
     modified_files = []
+    allowed_files = []
     for file in rsp.json().get("files", []):
         if file["status"] == "added":
-            added_files_names.append(file["filename"])
+            added_files.append(file["filename"])
         else:
             # To prevent the modifications to previously merged bundles,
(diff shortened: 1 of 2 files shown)


operator-pipeline-images/operatorcert/__init__.py
```diff
@@ -221,7 +221,7 @@ def get_files_added_in_pr(
     added_files_names = []
     modified_files = []
     for file in rsp.json().get("files", []):
-        if file["status"] == "added":
+        if file["status"] == "added" or file["filename"].endswith("ci.yaml"):
```
amisstea, line 224: I suppose this is ok but it's a little confusing since we want to allow `ci.yaml` to be _modified_. Would it be better to allow `ci.yaml` to be appended to `modified_files` and add an exception (with a clear log message) further down? I'm not sure of the implications of excluding it from the returned list in `added_files_names` though.

amisstea:
Please include unit tests to get the code coverage up to 100%.

amisstea (approved):

amisstea:
Closing in favor of #176.

PR #172, opened 2021-11-15
One role for both marketplace and certified operators

MarcinGinszt:
@amisstea I know we have discussed creating separate rolebindings for the marketplace and certified operators. However, I think simply creating a role per imagestream will be sufficient for us- roles should no longer override themselves, and all the imagestreams should have sufficient permissions. 
This also resolves another issue- as we were using vendor label to create the name of the role, one partner supplying two operators would also make the roles override themselves.


amisstea (approved):
Looks great! This is exactly what I had in mind :smile:

PR #178, opened 2021-11-17
Fix relatedImages bug and improve err msg

tonytcampbell:
Missed a bug on this that was referencing the wrong variable. We were using `$REPLACEMENT_COUNT` instead of `$REFERENCE_COUNT` in the `verified_related_images` step. 

Fixing case were the digest tool places duplicate images in the `relatedImages `section.  Now we check that the number of images in the `relatedImages` section is greater than or equal to the `$REFERENCE_COUNT`

Also improved the error message.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
@@ -99,10 +99,10 @@ spec:
         CSVFILE=$(find $BUNDLE_PATH -name "*clusterserviceversion.yaml" -o -name "*clusterserviceversion.yml")
         RELATED_IMAGE_COUNT=$(yq -e '.spec.relatedImages | length' $CSVFILE)
 
-        if [[ $REPLACEMENT_COUNT == $RELATED_IMAGE_COUNT ]]; then
+        if [[ $RELATED_IMAGE_COUNT >= $REFERENCE_COUNT ]]; then
           echo "Related images section exists."
           echo -n "true" | tee $(results.related_images_flag.path)
         else
-          echo "Related images section does not exist."
+          echo "The relatedImages section does not exist or is missing images."
           echo -n "false" | tee $(results.related_images_flag.path)
         fi


amisstea (approved):

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
```diff
@@ -99,10 +99,10 @@ spec:
         CSVFILE=$(find $BUNDLE_PATH -name "*clusterserviceversion.yaml" -o -name "*clusterserviceversion.yml")
         RELATED_IMAGE_COUNT=$(yq -e '.spec.relatedImages | length' $CSVFILE)
 
-        if [[ $REPLACEMENT_COUNT == $RELATED_IMAGE_COUNT ]]; then
+        if [[ $RELATED_IMAGE_COUNT >= $REFERENCE_COUNT ]]; then
```
MarcinGinszt, line 102: What are the contents of the file references.json? I guess it's different than replacements.json, but it's not documented anywhere, not even in the repository https://github.com/containerbuildsystem/operator-manifest.
Can the length of this file equal 0 for the proper submission? If yes, then we won't be able to validate, whether the key `.spec.relatedImages` exists.

vboulineau:
I think we're hitting this bug while trying to publish: https://github.com/redhat-openshift-ecosystem/redhat-marketplace-operators/pull/17

BTW, it would be nice to document how to run the `pinning` tool manually (running `operator-manifest pin <folder>` from `quay.io/redhat-isv/digest-pinning-tool`) as some transformations are basically impossible to guess.

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
amisstea: @lcarva, @MartinBasti would it be possible to update the `operator-manifest` docs with some info about this `references.json` file or does that already exist?

amisstea:
> I think we're hitting this bug while trying to publish: [redhat-openshift-ecosystem/redhat-marketplace-operators#17](https://github.com/redhat-openshift-ecosystem/redhat-marketplace-operators/pull/17)
> 
> BTW, it would be nice to document how to run the `pinning` tool manually (running `operator-manifest pin <folder>` from `quay.io/redhat-isv/digest-pinning-tool`) as some transformations are basically impossible to guess.

tonytcampbell:
> I think we're hitting this bug while trying to publish: [redhat-openshift-ecosystem/redhat-marketplace-operators#17](https://github.com/redhat-openshift-ecosystem/redhat-marketplace-operators/pull/17)
> 
> BTW, it would be nice to document how to run the `pinning` tool manually (running `operator-manifest pin <folder>` from `quay.io/redhat-isv/digest-pinning-tool`) as some transformations are basically impossible to guess.

Thanks @vboulineau we will add some documentation around this. This fix makes it so you don't have to match the pinning tool exactly.....but I agree the docs would be helpful as well.

ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
lcarva: It's just an array of strings, e.g.:
```
[
  "registry.example.com/org/spam:latest",
  "registry.example.com/org/ham:1.2"
]
```

MartinBasti: It's just in CLI help, so example wouldn't hurt. I can add it there.

```
operator-manifest replace --help

...

  REPLACEMENTS_FILE  The path to the replacements file. The format of this file is a simple JSON object where each attribute is a string representing the original image reference and the value is a string representing the new value for
                     the image reference. Use - to specify stdin.
```

MartinBasti: https://github.com/containerbuildsystem/operator-manifest/pull/44

PR #181, opened 2021-11-19
Use secretKeyRef instead of passing secrets via workspaces

wcheang:


wcheang:
I went through all the tasks to identify any env vars that need to be set, so there shouldn't be any more task that use echo to set a secret as an env var. However, there are still some tasks that have commands calls involving secrets, so those tasks still contain risks of leaking secrets and can't use `set -x`. I haven't thought of a good way to mitigate them from the brief research I did, and they're outside the scope of ISV-1278 anyway. If needed I can file a follow-up.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
```diff
@@ -41,15 +41,19 @@ spec:
     - name: pipeline_image
       description: An image of operator-pipeline-images.
       default: "quay.io/redhat-isv/operator-pipelines-images:v1.0.7"
+    - name: pyxis_api_key_secret_name
+      description: Kubernetes secret name that contains Pyxis API key. Valid only when external Pyxis is used.
+      default: pyxis-api-secret
```
wcheang, line 46: The default values of `pyxis_api_key_secret_name` and `pyxis_api_key_secret_key` can't be empty strings, because there are specific naming requirements for secret names that an empty string can't fulfill, and would cause an error.

README.md
```diff
@@ -119,14 +121,16 @@ tkn pipeline start operator-hosted-pipeline \
   --param env=prod \
   --param preflight_min_version=0.0.0 \
   --param ci_min_version=0.0.0 \
+  --param hydra_secret_name=hydra-credentials \
+  --param hydra_secret_username_key=username \
+  --param hydra_secret_password_key=password \
```
amisstea, line 126: In the interest of keeping these instructions clean and easier to get started with, I suggest we don't specify params for which the default value follows the setup instructions. Now that we have the ability to deploy our own custom envs, only a power user would need to mess with these. Wdyt?

README.md
```diff
@@ -119,14 +121,16 @@ tkn pipeline start operator-hosted-pipeline \
...
   --workspace name=repository,volumeClaimTemplateFile=templates/workspace-template-small.yml \
   --workspace name=results,volumeClaimTemplateFile=templates/workspace-template.yml \
   --workspace name=registry-credentials-all,volumeClaimTemplateFile=templates/workspace-template-small.yml \
   --workspace name=registry-credentials,secret=registry-dockerconfig-secret \
   --workspace name=pyxis-ssl-credentials,secret=operator-pipeline-api-certs \
   --workspace name=prow-kubeconfig,secret=prow-kubeconfig \
   --workspace name=preflight-decryption-key,secret=preflight-decryption-key \
```
amisstea, line 133: What about all these other workspaces from secrets? Are you planning to submit additional pull requests for the rest? They all seem like they can be eliminated from the list of required `tkn` args.

amisstea (changes requested):

README.md
wcheang: Sounds good. They're not really needed and does clutter up the command quite a bit. Updated.

README.md
wcheang: Oh I thought the task only involved ones that we had to export. Yeah I'll work on a follow-up PR to take care of the rest, can we merge this first to minimize conflicts?

amisstea: Since this includes an update to the tkn command for the CI pipeline, it'll affect partners who pull from our `main` branch. That requires partner-facing documentation updates so it'd be ideal to try and merge all the CI pipeline changes together.

wcheang: Okay sure. The conflicts should be relatively easy to deal with anyway.

wcheang: I looked over the rest of the secret workspaces, and I don't think setting them as env var the same way would make sense. The remaining secrets are secret files, which gets passed to commands/scripts as file paths. But if we convert them to use `secretKeyRef`, then the env var will be saving content of the file, which we'll then have to echo into a file, which actually creates the security risk we were trying to avoid in the first place.

However, it looks like we could maybe mount the secret files as a volume instead? Should I pursue that? e.g. https://kubernetes.io/docs/concepts/configuration/secret/#use-case-pod-with-ssh-keys

amisstea: Yes, please investigate the volume mount option further.

wcheang: Using volumeMount seems to work fine. But just to be clear, this will not change anything with the security. It'll just be for getting rid of the workspaces. We're still referencing files (except from a volumeMount instead of a workspace) and some tasks still need to `cat` the files and thus cannot use `set -x`

amisstea: Ok that makes sense :+1:. Anything we can do to reduce the amount of required params is welcome.

ansible/roles/operator-pipeline/templates/openshift/tasks/merge-registry-credentials.yml
```diff
@@ -37,8 +53,7 @@ spec:
           echo "Project contains secret registry tokens. Merging with internal tokens..."
           echo $PROJECT_CONFIG_JSON | sed 's/\\n/\n/g' > project_dockerconfigjson.gpg
 
-          gpg -v --batch --import <(cat "$(workspaces.gpg-key.path)/exd-guild-isv.key")
-          GPG_PASSPHRASE=`(cat "$(workspaces.gpg-key.path)/exd-guild-isv.passphrase")`
+          gpg -v --batch --import <$GPG_KEY
```
wcheang, line 56: @Allda I had a bit of trouble wrapping my head about the original line (`gpg -v --batch --import <(cat "$(workspaces.gpg-key.path)/exd-guild-isv.key")`, could you please take a look at this and verify that it's updated correctly? Thanks.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-cert-project-related-data.yml
```diff
@@ -33,19 +34,26 @@ spec:
...
+        secretName: "$(params.pyxis_ssl_secret_name)"
+        optional: true
   steps:
     - name: get-cert-project-related-data
       image: "$(params.pipeline_image)"
       workingDir: $(workspaces.source.path)
       env:
         - name: PYXIS_CERT_PATH
-          value: $(params.pyxis_cert_path)
+          value: /etc/pyxis-ssl-volume/$(params.pyxis_ssl_cert_secret_key)
```
Allda, line 50: Python scripts that talk to Pyxis use API key or Cert auth based on if a cert path is set. In the old code, the default value for cert path was set to `""` which caused the script used an API key. Now when you always set the cert path how does the script know when to use a cert and when to use API key?

ansible/roles/operator-pipeline/templates/openshift/tasks/merge-registry-credentials.yml
Allda: I think it looks ok. Did you verify it works?

wcheang: So it turned out it didn't work :( I updated it to a format much closer to the original, and it ran okay.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-cert-project-related-data.yml
wcheang: This was fixed in the latest pipeline image with a check for valid filepaths.

Allda (approved):

haripate (approved):
LGTM

PR #185, opened 2021-11-22
Parameterize job url to get artifacts

mrhillsman:

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
@@ -143,10 +143,16 @@ spec:
             exit 0
           fi
 
+          if [ "$(params.ocp_version)" == "4.6" ] || [ "$(params.ocp_version)" == "4.7" ]; then
+              JOB_SUFFIX="aws"
+          else
+              JOB_SUFFIX="claim"
+          fi
+
           BASE_URL="https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/"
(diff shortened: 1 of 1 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
```diff
@@ -143,10 +143,16 @@ spec:
             exit 0
           fi
 
+          if [ "$(params.ocp_version)" == "4.6" ] || [ "$(params.ocp_version)" == "4.7" ]; then
```
Allda, line 146: The same code already exists in the previous step. Can you please de-duplicate it?

tonytcampbell: @Allda This is a common pattern we've used in many task/steps. Each step runs in its own container so I think this would need to be checked in each step as written.  Do you have a suggestion for a better pattern?

mrhillsman: @Allda we cannot de-duplicate this unless you have a suggested way. My understanding is that we cannot easily pass environment variables between tasks. See the following as an example of this pattern:

```shell
if [ "$(params.skip)" == "true" ]; then
    echo "Preflight testing has been skipped"
    exit 0
fi
```

Allda: You can use a shared workspace (emptyDir?) to pass data from one step to another and it would de-duplicate a code.

I am fine with merging the PR as is if it is a blocking issue but a duplicated code is always an issue for maintenance in a long term.

tonytcampbell: ack....If we can push this to prod, we will raise an issue to address this tech debt. @Allda

Allda (approved):

PR #187, opened 2021-11-22
Verify marketplace annotation

Allda:
This commit adds a marketplace specific validation that checks required
marketplace annotation. The validation is done in ci and hosted
pipelines.

This commit also creates new tasks that were originally part of
different tasks. This is a refactor to avoid duplicated code an
unnecessary API calls.

JIRA: ISV-1366


MarcinGinszt (approved):
LGTM

PR #190, opened 2021-11-23
[ISV-1280] Add developer flag to enable/disable the query-publishing-checklist task

haripate:


haripate:
@Allda @MarcinGinszt Please approve or disapprove the current change, also if you have a better solution in mind please give your suggestions.

JIRA: https://issues.redhat.com/browse/ISV-1280

operator-pipeline-images/operatorcert/entrypoints/hydra_checklist.py
```diff
@@ -74,7 +81,7 @@ def main() -> None:
         log_level = "DEBUG"
     logging.basicConfig(level=log_level)
 
-    check_hydra_checklist_status(args.cert_project_id, args.hydra_url)
+    check_hydra_checklist_status(args.cert_project_id, args.hydra_url,args.developer)
```
MarcinGinszt, line 84: Hey! Current solution looks nice but it won't work as intended, as the `args.developer` passed from the pipeline to script is a string and not bool- so this step will always be skipped. Please, next time test as much of variants as possible for your features. 
To make things simpler, maybe we could just not trigger `hydra-checklist` script in case flag is set to true- the same way as it happens [here](https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml#L42)?

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -41,6 +41,9 @@ spec:
       default: quay-oauth-token
     - name: quay_oauth_secret_key
       default: token
+    - name: developer
```
MarcinGinszt, line 44: This flag is not too self- explanatory- what do you think about naming it something like `skip_hydra_check`?

operator-pipeline-images/operatorcert/entrypoints/hydra_checklist.py
haripate: Yes, I know this fact that with the current setup the code won't work. I just added it to ask the team for suggestions, the PR is still in draft. And according to the [AC], the hydra-checklist should be triggered even after the flag, the task is to prevent the failure while the developer is running the pipeline for testing purposes. WDYT?

Allda (changes requested):
I don't think you need to do any changes to a python script at all. Just check a hydra disable param in tekton task and skip a script completely.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
amisstea: I'd make it even more descriptive. Hydra is quite a large service with many components. How about `ignore_publishing_checklist` instead? We also don't want to skip the query but rather ignore the results.

operator-pipeline-images/operatorcert/entrypoints/hydra_checklist.py
amisstea: We should still query the publishing checklist because there is value in testing its availability and correctness of the response. For testing purposes though, it's nice to be able to ignore the `completed` state.

MarcinGinszt: I though, that we want to skip the task completelly- if we still want to query Hydra, I agree with Alex.

haripate:
UPDATE: I have modified the `main-image.yml` task just for temporary testing purposes. Before merging this PR I will undo the changes.

haripate:
The code has been tested and verified in the test cases, it is working as expected. 

cc @amisstea @Allda @MarcinGinszt Please review it now.

amisstea (commented):
Please also add a section about this feature to the usage docs for the hosted pipeline in `README.md` (rebase first).

ansible/roles/operator-pipeline/templates/openshift/tasks/query-publishing-checklist.yml
```diff
@@ -32,4 +35,5 @@ spec:
 
         hydra-checklist \
           --cert-project-id "$(params.cert_project_id)" \
-          --hydra-url $HYDRA_URL
+          --hydra-url $HYDRA_URL \
+          --ignore-publishing-checklist "$(params.ignore_publishing_checklist)"
```
Allda, line 39: Is this really necessary to pass the flag to the python script? If we want to just ignore a checklist why don't you just not call the python script? 
By skipping the python script you don't have to do any changes in the script at all.

amisstea: I requested this in the acceptance criteria.

> The checklist should still be queried to ensure the configuration is correct.

amisstea (approved):

PR #195, opened 2021-12-02
Limit the version of ocp passed to preflight-trigger

mrhillsman:
When a version of ocp is listed as supported by an operator (4.10) but is not available in DPTP (max 4.9) we should ensure that a request for a 4.9 cluster is made.


tonytcampbell (approved):
/lgtm

Allda (approved):

PR #198, opened 2021-12-08
Rotate ISV gpg keys

Allda:
The new set of keys has been created for prod and non-prod environments
due to a recent leak.

JIRA: ISV-1470


amisstea (approved):
LGTM :+1:

PR #199, opened 2021-12-09
PPC64LE image tags should have prefix

MarcinGinszt:
Otherwise they overwrite our, existing images.
ISV-1495

### DIFF
--- a/.github/workflows/ppc64le-image.yml
+++ b/.github/workflows/ppc64le-image.yml
@@ -21,7 +21,7 @@ jobs:
         uses: redhat-actions/buildah-build@v2
         with:
           image: operator-pipelines-images
-          tags: ppc64le ${{ github.sha }}
+          tags: ppc64le ppc64le-${{ github.sha }}
           archs: ppc64le
           dockerfiles: |
             ./operator-pipeline-images/Dockerfile-ppc64le
--- a/.github/workflows/ppc64le-image.yml
+++ b/.github/workflows/ppc64le-image.yml
@@ -21,7 +21,7 @@ jobs:
         uses: redhat-actions/buildah-build@v2
         with:
           image: operator-pipelines-images
-          tags: ppc64le ppc64le-${{ github.sha }}
+          tags: ppc64le-${{ github.sha }}
           archs: ppc64le
           dockerfiles: |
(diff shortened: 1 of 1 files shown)


MarcinGinszt:
cc: @mayurwaghmode

.github/workflows/ppc64le-image.yml
```diff
@@ -21,7 +21,7 @@ jobs:
         uses: redhat-actions/buildah-build@v2
         with:
           image: operator-pipelines-images
-          tags: ppc64le ${{ github.sha }}
+          tags: ppc64le-${{ github.sha }}
```
Allda, line 24: This change has the side effect of removing a floating tag from ppc64le images. Could you keep ppc64le tag and add a separate one `ppc64le-sha`?

MarcinGinszt: Floating tag is pointless, as long as the ppc64le image is built on every commit. WDYT?

MarcinGinszt: @mayurwaghmode what should we do with this tag?

mayurwaghmode: @MarcinGinszt Can we have a tag as ppc64le:latest?

MarcinGinszt: Yes, we can. But would you like to build it for every commit on every branch, or maybe only for master?

rpsene: what if we set the ppc64le-$BRANCH-${{ github.sha }} for every branch and ppc64le:$RELEASE for every new release?

MarcinGinszt: @rpsene @mayurwaghmode  You are encouraged to submit the PR in order to create a custom image tagging system. I just want to unblock us by removing the bug, that overrides images necessary for prod.

MarcinGinszt: @rpsene @mayurwaghmode  can we merge this one?

rpsene: @MarcinGinszt for now yes. We will update it later.

Allda (approved):

PR #206, opened 2021-12-15
Using proxy for preprod environments

MarcinGinszt:


Allda:
@MarcinGinszt There are tasks that use Pyxis API using python scripts. If I understand this change correctly the proxy will not work for these tasks. How does that work?

Allda (approved):

PR #209, opened 2021-12-17
Configure CI/CD to use a latest image build

Allda:
The deploy action was extended with a build configuration and the
individual steps of release use the image with the latest commit sha.

This pull request includes several changes:
 - Build container image before releasing it to env
 - Use commit sha tag when deploying pipelines
 - Create GitHub release including new Github tag and container repository sem-ver tag
 - Remove duplicated action runs
 - Remove artifacts that stayed in a repo when pipeline repo was merged

JIRA: ISV-1363, ISV-1364


.github/workflows/validation.yml
```diff
@@ -2,13 +2,25 @@
 name: validation
 
 on:  # yamllint disable-line rule:truthy
-  - push
-  - pull_request
+  pull_request:
+  push:
```
wcheang, line 6: Don't we want the validation to run on all pushes, in case people make changes while the pull request is open?

wcheang:
Just one question, the rest looks good.

.github/workflows/validation.yml
Allda: The `pull_request` by default also reacts on push events, so if you have PR and push a new code into the branch the workflow is triggered.

The change in workflow triggers should remain the same from our PoV but at the same time, it de-duplicates actions that are triggered by both PR and push events.

https://frontside.com/blog/2020-05-26-github-actions-pull_request/

PR #214, opened 2022-01-04
Extract build step to separate action

Allda:
The build action is used by both e2e and deployment action and is
executed only once.


wcheang (approved):

Allda:
ISV-1542

PR #217, opened 2022-01-10
[ISV-1353] The hosted/release pipelines should include links to the results/logs in a PR comment

haripate:


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -460,6 +460,25 @@ spec:
         - name: GITHUB_TOKEN_SECRET_KEY
           value: "$(params.github_token_secret_key)"
 
+    # Add the connect page URL in github PR comment to view the test results/logs
+    - name: github-add-connect-url-comment
```
wcheang, line 464: Wouldn't this lead to two identical comments posted, after both hosted and release pipelines are run?

wcheang:
The rest looks good, I just have the one question.

docs/developer-guide.md
```diff
@@ -118,7 +118,7 @@ The pipelines depend on the following certification project fields:
   "operator_distribution": "<insert-distribution>",
 
   // Must correspond to a containerVendor record with the same org_id value.
-  "org_id": <insert-org-id>,
+  "org_id": "<insert-org-id>",
```
amisstea, line 121: This field is an int, not a str so the quotes don't help.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
amisstea: It seems so. It'd be nicer to have distinct comments. For example, a user shouldn't care to see test results after the release pipeline succeeds or fails. If this pipeline is running they've already moved past certification but they may still wish to look at the release pipeline logs.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -846,46 +846,31 @@ spec:
...
-    - name: set-github-status-success
+    # Add the connect page URL in github PR comment to view the test results/logs
+    - name: github-add-connect-url-comment
       when:
-        - input: "$(tasks.status)"
+        - input: "$(tasks.certification-project-check.status)"
           operator: in
-          values: ["Succeeded", "Completed"]
+          values:
+            - Succeeded
```
amisstea, line 855: I don't think we need this `when` condition. Tekton should be smart enough to notice the dependency on `certification-project-check` and skip this task if the other didn't succeed.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -460,6 +460,25 @@ spec:
...
+      when:
+        - input: "$(tasks.certification-project-check.status)"
+          operator: in
+          values:
+            - Succeeded
```
amisstea, line 469: See my other comment in the hosted pipeline. I think we can remove this `when` condition.

amisstea (changes requested):

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
haripate: As per my understanding, we need to add the two comments for the same link anyway, just need to distinguish the comment message. If this is what we wanted, then what if we just add the prefix `hosted or release` pipeline in the comment message?

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
haripate: Yes, it seems no use of when condition here. I will remove it.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
haripate: Same as above, I will remove the when condition here too.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
amisstea: Or maybe we drop this comment from the release pipeline altogether. It doesn't seem to add any value and the message at the moment is generic enough for all use cases. Wdyt?

haripate: I think it would be better to just drop the message from the Release pipeline.

wcheang: Agreed. No point posting the same link again.

haripate: Okay, then I will remove the comment from the release pipeline.

wcheang (approved):

amisstea (approved):

PR #219, opened 2022-01-11
Fill container image image_id

Allda:
Operator pipelines were missing a field image_id that is used by some
Pyxis endpoints. This commit adds image_id into container image creation
payload.


haripate (approved):
LGTM

amisstea (approved):

PR #225, opened 2022-01-17
Add retries for all HTTP(S) requests made with requests

amisstea:
The retry status codes are a best effort guess at this point but should
be sane defaults to start with. The same goes for the backoff strategy.

### DIFF
--- a/operator-pipeline-images/operatorcert/github.py
+++ b/operator-pipeline-images/operatorcert/github.py
@@ -3,8 +3,8 @@ import os
 from typing import Any, Dict, Optional
 
 import requests
-from requests.adapters import HTTPAdapter
-from requests.packages.urllib3.util.retry import Retry
+
+from operatorcert.utils import add_session_retries
 
 LOGGER = logging.getLogger("operator-cert")
 
@@ -25,14 +25,7 @@ def _get_session(auth_required: bool = False) -> requests.Session:
         requests.Session: Github session
     """
     session = requests.Session()
-
-    # Exponential retry backoff for a max wait of ~8.5 mins
-    retries = Retry(
-        total=10, backoff_factor=1, status_forcelist=(408, 500, 502, 503, 504)
-    )
-    adapter = HTTPAdapter(max_retries=retries)
-    session.mount("http://", adapter)
-    session.mount("https://", adapter)
+    add_session_retries(session)
 
     if auth_required:
         token = os.environ.get("GITHUB_TOKEN")
--- a/operator-pipeline-images/operatorcert/github.py
+++ b/operator-pipeline-images/operatorcert/github.py
@@ -1,8 +1,10 @@
 import logging
 import os
-from typing import Any, Dict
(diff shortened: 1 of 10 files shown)


wcheang (approved):

haripate (approved):
LGTM

operator-pipeline-images/operatorcert/github.py
```diff
@@ -22,18 +24,62 @@ def _get_session(auth_required: bool = False) -> requests.Session:
...
+        Dict[str, Any]: GitHub response
+    """
+    session = _get_session(auth_required=auth_required)
+    LOGGER.debug(f"GET GitHub request url: {url}")
+    LOGGER.debug(f"GET GitHub request params: {params}")
+    resp = session.get(url, params=params)
+
+    try:
+        resp.raise_for_status()
+    except requests.HTTPError:
```
Allda, line 74: Just wanna make sure. Does this also cover the MaxRetry exception?

operator-pipeline-images/operatorcert/hydra.py
```diff
@@ -28,6 +31,15 @@ def _get_session() -> requests.Session:
 
     session = requests.Session()
     session.auth = (username, password)
+
+    # Exponential retry backoff for a max wait of ~8.5 mins
+    retries = Retry(
```
Allda, line 36: This is used in several places. Can you please extract it into a separate module or function and reuse it everywhere else?

operator-pipeline-images/operatorcert/github.py
amisstea: It doesn't. That error can be raised by `session.get` so will not necessarily be caught by this exception handler anyway. Prior to this change, retries were still enabled but the delay between attempts was 0. Therefore, this shouldn't change the current behavior for retry exception handling.

operator-pipeline-images/operatorcert/hydra.py
amisstea: Sure, I'll move it into `utils.py`.

operator-pipeline-images/operatorcert/github.py
Allda: I raised this question because from the recent experience when the retries session reaches a maximum attempt it just raises the MaxRetry exception with no details about what went wrong.

It would be great to log HTTP errors with an error code also for Retry exceptions. If you don't have the capacity to do it now can you please file a ticket? Thanks

amisstea: I think I found a nice solution to this. Setting `raise_on_status=False` on the `Retry` object will stop `MaxRetryError` exceptions from being raised by urllib3 (at least when there is a response code). `raise_for_status` would then raise them normally.

amisstea: Please take another look when you get a chance. Thanks!

Allda (approved):

PR #227, opened 2022-01-19
[ISV-1612] Increase catalog image import timeout

amisstea:
Apparently 5 mins may not be enough (see https://github.com/redhat-openshift-ecosystem/operator-pipelines/runs/4874526765?check_suite_focus=true).


haripate (approved):
LGTM

PR #228, opened 2022-01-21
Fixed error for execution of the CI pipeline on non-amd64 cluster

mayurwaghmode:
Fixed below error at the digest-pinning step for Pipeline Run with Image Digest Pinning on the non-amd64 cluster -
PermissionError: [Errno 13] Permission denied: '/workspace/source/operators/ppc64le-cluster-testing-operator/1.12.0/manifests/ppc64le-cluster-testing-operator.clusterserviceversion.yaml'

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/git-clone.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/git-clone.yml
@@ -211,3 +211,4 @@ spec:
         fi
         printf "%s" "${RESULT_SHA}" > "$(results.commit.path)"
         printf "%s" "${PARAM_URL}" > "$(results.url.path)"
+        chmod -R 777 "${CHECKOUT_DIR}"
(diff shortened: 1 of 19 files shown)


mayurwaghmode:
@amisstea, @Allda can u please review?

ansible/roles/operator-pipeline/templates/openshift/tasks/git-clone.yml
```diff
@@ -211,3 +211,4 @@ spec:
         fi
         printf "%s" "${RESULT_SHA}" > "$(results.commit.path)"
         printf "%s" "${PARAM_URL}" > "$(results.url.path)"
+        chmod -R 777 "${CHECKOUT_DIR}"
```
amisstea, line 214: Does this dirty the git status? Please confirm if it works with digest pinning enabled. We need to ensure it won't produce a commit with every file permission changed.

amisstea:
We discovered that this isn't compatible with the git commit after automatic digest pinning because it touches every file. E.g.

```
[main-pinned 7dda390] Pin the operator manifests
 1656 files changed, 1215 insertions(+), 1210 deletions(-)
 ```

More investigation is needed so we've agreed to close this.

PR #229, opened 2022-01-22
Add task and script to upload signature to Pyxis sigstore

wcheang:


ansible/roles/operator-pipeline/templates/openshift/tasks/upload-signature.yml
```diff
@@ -0,0 +1,73 @@
...
+      image: "$(params.pipeline_image)"
+      script: |
+        #! /usr/bin/env bash
+        set -xe
+
+        if [[ "$(params.verify_signature)" == "true" ]]; then
+          echo "Verifying signature before upload"
+          echo $(params.signature_data) | base64 --decode > decoded_signed_claim
+
+          # import containerisvsign pub key
```
wcheang, line 44: There might be a better way to import the pub key, but trying to bake it into the container image didn't work (even putting it after the user change). If you can think of a better way please let me know.

ansible/roles/operator-pipeline/templates/openshift/tasks/upload-signature.yml
```diff
@@ -0,0 +1,73 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: upload-signature
+spec:
+  params:
+    - name: docker_reference
+    - name: manifest_digest
+    - name: repository
```
amisstea, line 10: Some descriptions would be nice to have here. Especially if they contain examples. Got anything?

ansible/roles/operator-pipeline/templates/openshift/tasks/upload-signature.yml
amisstea: One option is to download the key during the image build and set an environment variable to its path. Continue to run the `gpg` commands in this task but reference the env var. If it's possible, parametrize the env variable name in this task too.

Wdyt? Do you think that'll work?

wcheang: Yup, that works! Updated, thanks for the input.

ansible/roles/operator-pipeline/templates/openshift/tasks/upload-signature.yml
wcheang: For sure. Updated.

operator-pipeline-images/operatorcert/entrypoints/upload_signature.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    payload = {
+        "manifest_digest": args.manifest_digest,
+        "reference": args.reference,
+        "repository": args.repository,
+        "sig_key_id": args.sig_key_id,
+        "signature_data": args.signature_data,
+    }
+
+    rsp_json = pyxis.post(urljoin(args.pyxis_url, "v1/signatures"), payload)
```
Allda, line 73: In case we would need to re-trigger a pipeline this POST call would raise 409 code I guess. Can you maybe catch the 409 exception, log a warning and continue?

ansible/roles/operator-pipeline/templates/openshift/tasks/upload-signature.yml
```diff
@@ -0,0 +1,78 @@
...
+      description: Docker reference for the signed content, e.g. registry.redhat.io/redhat/community-operator-index:v4.9
+    - name: manifest_digest
+      description: Manifest digest for the signed content, usually in the format sha256:xxx
+    - name: repository
+      description: Name of the repository that hosts the signed content, e.g. redhat/community-operator-index
+    - name: sig_key_id
+      description: The signing key id that the content was signed with.
+    - name: signature_data
+      description: The signature to upload to Pyxis signature.
+    - name: pipeline_image
```
Allda, line 18: Please add a description.

operator-pipeline-images/operatorcert/entrypoints/upload_signature.py
wcheang: Good idea. Added.

Allda (approved):

PR #238, opened 2022-02-02
Fix logger usage in entrypoints

Allda:
2 scripts were using a default logger instead of the one that is used by
other scripts and pre-configured. This caused missing logs in these
scripts.

JIRA: ISV-1682


kosciCZ:
LGTM :+1:

kosciCZ (approved):

PR #243, opened 2022-02-04
Use environment specific registry when creating image

Allda:
The create image script now uses a environment specific registry based
on pipeline argument. Each environment uses a different registry while
creating container image metadata.

JIRA: ISV-1674


amisstea (approved):

haripate (approved):
LGTM

PR #247, opened 2022-02-09
Remove unused ppc64le build action

Allda:
The action was replaced by the multi-arch build action and it is no
longer needed.


Allda:
@mayurwaghmode just FYI based on what we discussed last week.

PR #248, opened 2022-02-10
[ISV-1678] Update Pyxis external URLs in set-env task

AdamJel:


ansible/roles/operator-pipeline/templates/openshift/tasks/set-env.yml
```diff
@@ -70,7 +70,7 @@ spec:
                     PYXIS_URL="https://pyxis.qa.engineering.redhat.com"
                     ;;
                     external)
-                    PYXIS_URL="https://pyxis-qa-ext.int.us-west.dc.preprod.paas.redhat.com"
+                    PYXIS_URL="https://catalog.qa.redhat.com/api/containers/v1"
```
Allda, line 73: Please remove the `/v1` suffix to align with stage and prod URLs.

PR #254, opened 2022-02-15
ISV-1711,timeout is increased to 1 hour from 30 min

atiwary91:


amisstea:
@atiwary91 please also update the timeout for the entire release pipeline. By default, a Tekton pipeline will timeout after 1h. Here's [an example](https://github.com/redhat-openshift-ecosystem/operator-pipelines/commit/f9f839821b856f6a834f9eec68c47beaf31ad610) of setting it on the hosted pipeline. Perhaps 1.5 hours would be good.

PR #258, opened 2022-02-22
Add task to request signing of an index image from RADAS

wcheang:


wcheang:
@Allda @haripate this is still a WIP (need unit tests etc.), but the essentials like input and output should be all there. No reviews needed now but feel free to reference it as needed for any of your signing pipeline tasks.

operator-pipeline-images/operatorcert/entrypoints/request_signature.py
```diff
@@ -0,0 +1,248 @@
...
+
+    return umb_client
+
+
+def process_message(frame: Any):
+    msg_dict = json.loads(frame.body)
+    msg = msg_dict["msg"]
+
+    global umb
+    if msg.get("request_id") == umb.id:
```
Allda, line 150: Do I understand it right that the task sends a signature request with `id` field and waits for responses with the same `request_id` to match a request with a response?

wcheang: Yes, the `request_id` is how we can tell which signature is for the request we sent, when there are multiple going on.

ansible/roles/operator-pipeline/templates/openshift/tasks/request-signature.yml
```diff
@@ -0,0 +1,76 @@
...
+    - name: umb_listen_topic
+      description: umb topic to listen to for responses with signed content
+      default: VirtualTopic.eng.robosignatory.isv.sign
+    - name: umb_publish_topic
+      description: umb topic to publish to for requesting signing
+      default: VirtualTopic.eng.operatorpipelines.isv.sign
+    - name: umb_url
+      description: umb host to connect to for messaging
+      default: umb.api.redhat.com
+  results:
```
Allda, line 38: Can you extend results to return values that are needed by upload-signature task?

Currently `repository` and `sig_key_id` are missing.

wcheang: `repository` wasn't actually used by the signing request, so I removed it from the params. I can add `sig_key_id`, but that's gonna need to be passed in from the pipeline based on environments (pre-prod should use the redhate2etesting key, key id 4096R/37036783 SHA-256

wcheang: If it makes things easier, I could set the sig_key_* within the request signing script instead, and take in an env param. I wanted to avoid hardcoding, but that's what RHC Cert Service does currently, and those values are very very unlikely to change.

Allda: If it requires something that depends on the environment you can add it into `set-env` task and I can re-use the value in the new pipeline and pass it to your task as an argument.

wcheang: Good point. Added all the env-dependent values to the `set-env` task in the latest commit.

Allda (approved):

operator-pipeline-images/operatorcert/entrypoints/request_signature.py
```diff
@@ -0,0 +1,252 @@
...
+    )
+    parser.add_argument("--verbose", action="store_true", help="Verbose output")
+    return parser
+
+
+global umb
+umb = None
+
+global result_file
+result_file = None
```
amisstea, line 94: I don't think you need to use the `global` keywords here.

amisstea (approved):
The rest LGTM :+1:

PR #268, opened 2022-03-01
[ISV-1493] use new ansible vault password for prod

mantomas:


amisstea (commented):
The changes look good. Since this is all related to the prod env, what testing have you done for it?

Allda (approved):

mantomas:
@amisstea I ran init-custom-env.sh against our openshift cluster using prod and stage env. Given the correct vault password both init finished successfully.

amisstea (approved):
LGTM :+1: @mantomas this is a risky change since we cannot test that the secrets continue to work without running a prod pipeline. Therefore, prior to merging, please double (or triple) check that all the decrypted secret values remain unchanged (watch out for newlines, etc.).

PR #269, opened 2022-03-01
[ISV-1661] Automate the import of index images using ansible

amisstea:
Each deployment will now create or update the index imagestreams
including any new tags. This also supports custom environments.


mantomas (approved):

Allda (approved):
That's awasome!

PR #272, opened 2022-03-08
[ISV-1662] Whitelist set of community index images that can be signed

haripate:


ansible/roles/operator-pipeline/tasks/webhook-event-listener.yml
```diff
@@ -101,6 +101,24 @@
                   - ref: operator-release-pipeline-trigger-binding
                 template:
                   ref: operator-release-pipeline-trigger-template
+              # run Community Signing pipeline only with whitelisted index images
```
Allda, line 104: @haripate This is defined in the wrong place. There is a separate webhook event for the community signing pipeline file. However, we should rather create a validation tekton task and add it into the signing pipeline.

haripate: We decided to add validation to the webhook event in the past sprint, but yes I can change it into a tekton task.

haripate: Removed the webhook trigger and added the tekton task to validate it.

Allda (approved):

PR #274, opened 2022-03-09
Update signing tasks to support both single and multiple requests

wcheang:


wcheang:
@Allda still finishing up testing, but could you please take a look at the new implementation and see if that's okay? I decided to change the input to upload-signature into a list of the signature data, which simplified the params a bit. There's minor impact to community signing pipeline that I've updated accordingly.

wcheang:
@Allda made some tweaks to save the request-signature output into a json file instead, since weird things happened with the space in `sig_key_id` when the output was just passed as a string. The new changes are in the second commit, please take another look. It's been tested with the community signing pipeline and worked fine.

Allda (approved):
Looks ok to me

operator-pipeline-images/operatorcert/entrypoints/request_signature.py
```diff
@@ -192,44 +203,92 @@ def request_signature(args: Any) -> None:
...
+                    if os.path.exists(f"{request_id}-{result_file}"):
+                        with open(f"{request_id}-{result_file}", "r") as f:
+                            result_json = json.load(f)
+                            if result_json["signing_status"] == "success":
+                                results.append(result_json)
+                                sig_received.add(request_id)
+                            else:
+                                LOGGER.error(
+                                    f"Signing failure received for request {request_id}"
+                                )
```
amisstea, line 270: Is this really a signing failure or could it just be pending?

operator-pipeline-images/operatorcert/entrypoints/request_signature.py
```diff
@@ -192,44 +203,92 @@ def request_signature(args: Any) -> None:
...
+
+                request_ids = request_ids - sig_received
             else:
-                # exit retry loop if response file detected
+                # exit retry loop if all response files detected
                 break
 
-            LOGGER.info(f"No signing response received. Retrying.")
+            LOGGER.info(f"Not all successful signing response received. Retrying.")
```
amisstea, line 277:
```suggestion
            LOGGER.info(f"Not all successful signing responses received. Retrying.")
```

operator-pipeline-images/operatorcert/entrypoints/request_signature.py
wcheang: So far the only two statuses for ISV signing are "success" and "failure". I can change the check to `!= "failure"` if we wanna be future-proof about adding more statuses.

amisstea: How about adding an `elif` condition to specifically check for the `failure` state? The `else` condition log message should then highlight the unknown state.

wcheang: Thanks, updated. Please check again.

amisstea (approved):

PR #280, opened 2022-03-14
[ISV-1821] Update publish-bundle to return list of references and manifest digests

wcheang:
The references and manifest digests returned from IIB results are
required for signing on the release pipeline.


wcheang:
I've tested this to the extend that I can, but IIB wasn't actually run because it doesn't run on dev/qa and it's broken in stage right now.

Allda (approved):

operator-pipeline-images/operatorcert/entrypoints/index.py
```diff
@@ -129,6 +139,27 @@ def publish_bundle(
...
+    response: Dict[str, Any],
+):
+
+    LOGGER.info("Extracting manifest digests for signing...")
+    manifest_digests = []
+    # go through each version to ensure order is the same as the indices list
+    for version in index_versions:
+        for build in response["items"]:
+            if build["from_index"].endswith(version):
+                digest = build["from_index_resolved"].split("@")[-1]
```
amisstea, line 158: Should we use `index_image` and `index_image_resolved` from the IIB response instead? The "from index" is the index prior to the addition of the bundle.

wcheang: Oh thanks for catching that. Fixed.

amisstea (approved):

PR #281, opened 2022-03-14
Add yamllint to tox

wcheang:
Adding yamllint to the tox run, so errors can be caught before the
github workflow.


Allda:
Yamllint is already part of Github validation. I am fine with moving it to Tox but can you also update validation GitHub action and add yamlint to tox section and remove the standalone yamllint action step?

PR #282, opened 2022-03-15
[ISV-1856] The Operator pipeline integration tests should always increment the bundle semver.

haripate:


.github/workflows/e2e.yml
```diff
@@ -29,10 +31,10 @@ jobs:
     steps:
       - id: prepare
         run: |
-          echo "::set-output name=short_sha::${GITHUB_SHA::7}"
-          echo "::set-output name=pr_title::operator test-e2e-operator (0.0.7-${GITHUB_SHA::7})"
-          echo "::set-output name=oc_project::test-e2e-${GITHUB_SHA::7}"
-          echo "::set-output name=test_pr_base_branch::${GITHUB_SHA::7}-base"
+          echo "::set-output name=short_sha::${{ github.run_id }}"
```
Allda, line 34: The description of run_id says: "A unique number for each workflow run within a repository. This number does not change if you re-run the workflow run." I think we can face the same issue if we re-run the workflow.

Allda: Also if you use a github id we should not call the variable `short_sha`

amisstea, line 34: Can you please rename this variable since it isn't a SHA?

haripate: I was also thinking the same as your first comment, and I came up with the solution to add `github.run_id` along with `github.run_number`. I will update it as well as the variable's name in my next commit.

amisstea: `github.run_attempt` might be more suitable in combination with either `github.run_id` or `github.run_number`. From the [docs](https://docs.github.com/en/actions/learn-github-actions/contexts):

 `gitlab.run_id`
> A unique number for each workflow run within a repository. This number does not change if you re-run the workflow run.

 `gitlab.run_number`
> A unique number for each run of a particular workflow in a repository. This number begins at 1 for the workflow's first run, and increments with each new run. This number does not change if you re-run the workflow run.

`github.run_attempt`
> A unique number for each attempt of a particular workflow run in a repository. This number begins at 1 for the workflow run's first attempt, and increments with each re-run.

haripate: Oh yes, it was `run_attempt`. Sorry for the confusion.

haripate:
**Note:** I have used the `Pull request number` rather than `github run_id` to represent the `version_prefix` because the long run_id value will be a cause in character limit excess for GitHub branch name we use in purpose to prepare the E2E testing as a parameter in the future. Also, I verified that if we re-run the E2E testing with the same version it doesn't cause a failure, so I avoid using the unique version semvers for the same commit.

Allda:
@haripate The e2e usually runs after the PR is already merged. Are you sure the PR number will work as a version?

.github/workflows/e2e.yml
```diff
@@ -22,17 +24,17 @@ jobs:
   prepare-env-job:
     runs-on: ubuntu-latest
     outputs:
-      short_sha: ${{ steps.prepare.outputs.short_sha }}
+      version_prefix: ${{ steps.prepare.outputs.version_prefix }}
```
Allda, line 27: This is more like postfix than prefix.

haripate:
> @haripate The e2e usually runs after the PR is already merged. Are you sure the PR number will work as a version?

- I think it should work even after merging the PR, but to avoid confusion and make it still efficient I used `github.run_number` as it is a unique number as same as `github.run_id`.

.github/workflows/e2e.yml
haripate: Updated.

haripate:
> > > @haripate The e2e usually runs after the PR is already merged. Are you sure the PR number will work as a version?
> > 
> > 
> > ```
> > * I think it should work even after merging the PR, but to avoid confusion and make it still efficient I used `github.run_number` as it is a unique number as same as `github.run_id`.
> > ```
> 
> With `github.run_number` I think we can face the same issues as I described in the first comment. See the description of the variable: "A unique number for each run of a particular workflow in a repository. This number begins at 1 for the workflow's first run, and increments with each new run. This number does not change if you re-run the workflow run ". The fact that the value remains the same if we re-run the workflow might cause the same issue.

- Yes, I added the `run_attempt` to the version_suffix to prevent any future failure.

.github/workflows/e2e.yml
```diff
@@ -22,17 +22,17 @@ jobs:
...
       - id: prepare
         run: |
-          echo "::set-output name=short_sha::${GITHUB_SHA::7}"
-          echo "::set-output name=pr_title::operator test-e2e-operator (0.0.7-${GITHUB_SHA::7})"
-          echo "::set-output name=oc_project::test-e2e-${GITHUB_SHA::7}"
-          echo "::set-output name=test_pr_base_branch::${GITHUB_SHA::7}-base"
+          echo "::set-output name=version_suffix::${{ github.run_number }}${{ github.run_attempt }}"
+          echo "::set-output name=pr_title::operator test-e2e-operator (0.0.8-${{ github.run_number }}${{ github.run_attempt }})"
+          echo "::set-output name=oc_project::test-e2e-${{ github.run_number }}${{ github.run_attempt }}"
+          echo "::set-output name=test_pr_base_branch::${{ github.run_number }}${{ github.run_attempt }}-base"
```
amisstea, line 35: According to the docs, both of these values start at 1 and increment. So isn't this problematic when there are 10 or more run attempts? E.g.

```bash
0.0.8-11   # Run 1, attempt 1
0.0.8-12   # Run 1, attempt 2
0.0.8-13   # Run 1, attempt 3
...
0.0.8-110  # Run 1, attempt 10
0.0.8-21   # Run 2, attempt 1
```

When sorted by semver, 0.0.8-110 is going to be considered a replacement of 0.0.8-21 by `opm`.

haripate: I don't think so. Because when I started using `run_number` instead of `run_id` and the opm worked fine at that point even though the `run_id` was a higher number than `run_number`.

amisstea: Adding a dash may even be sufficient (according to [this](https://semvercompare.azurewebsites.net) tool).

```bash
0.0.8-1-1   # Run 1, attempt 1
...
0.0.8-1-10  # Run 1, attempt 10
0.0.8-2-1   # Run 2, attempt 1
```

amisstea: Nevermind; adding a dash isn't good enough. As soon as the number of digits in the `run_number` goes above 9 it won't sort as desired.

haripate: Here is the suggestion: `0.0.{run_number}-{run_attempt}`. It always be increasing.

amisstea: That works well! Our run numbers are already much higher that 7 or 8. It shouldn't cause any conflicts with the stage index images.

.github/workflows/e2e.yml
```diff
@@ -151,11 +156,11 @@ jobs:
...
             --param git_repo_url=git@github.com:redhat-openshift-ecosystem/operator-pipelines-test.git \
             --param upstream_repo_name=redhat-openshift-ecosystem/operator-pipelines-test \
-            --param git_branch=$SHORT_SHA \
-            --param bundle_path=operators/test-e2e-operator/0.0.7-$SHORT_SHA \
+            --param git_branch=$VERSION \
+            --param bundle_path=operators/test-e2e-operator/$VERSION \
             --param env=stage \
             --param submit=false \
-            --param pipeline_image=quay.io/redhat-isv/operator-pipelines-images:${{ github.sha }} \
+            --param pipeline_image=quay.io/redhat-isv/operator-pipelines-images:latest \
```
amisstea, line 163: Do you intend to revert this change in all three pipeline start commands?

haripate: It is for testing purposes only. I will change it before merging the PR.

amisstea (approved):

PR #283, opened 2022-03-15
WIP add signing tasks to release pipeline

wcheang:


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -70,9 +73,15 @@ spec:
       description: >-
         The key within the Kubernetes Secret that contains the kerberos keytab for submitting IIB builds.
       default: krb5.keytab
-    - name: metrics_endpoint
-      description: Prometheus metrics endpoint
-      default: http://pipeline-metrics.pipeline-metrics-prod
+    - name: umb_ssl_secret_name
```
Allda, line 76: Is this required? UMB is using the same cert as Puxis. Can we just reuse it?

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -417,10 +426,70 @@ spec:
...
+        name: request-signature
+      params:
+        - name: pipeline_image
+          value: "$(params.pipeline_image)"
+        - name: manifest_digest
+          value: "$(tasks.publish-bundle.results.manifest_digests)"
+        - name: reference
+          value: "$(tasks.publish-bundle.results.docker_references)"
+        - name: requester
+          value: "amisstea"
```
Allda, line 443: Shouldn't this be an email?

amisstea: If it must be a username, could we use a service account backed by ldap?

Allda:
@wcheang I just realized that this PR needs to be released in sync with Radas changes and community certification changes. So please don't merge it now. We got a confirmation from community folks that they will be able to test preprod signing pipeline next week and then we have to synchronize the production release of all components including this PR.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
wcheang: We can reuse it sure. I was thinking it'll give some flexibility if in the future a different cert is used, but that's unlikely and we can always add the params in the future as needed. Updated

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
wcheang: So is this okay as is? @Allda @amisstea

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
wcheang: Updated for real, please check again.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
amisstea: I'm ok with it. I'll also document this in the guild. The guild master should be the one requesting signatures for as long as this is a requirement.

PR #289, opened 2022-03-23
Remove unused suffix from image tag

Allda:
When creating container image in Pyxis, the script was adding
unnecessary suffix "-1". This caused inaccurate pull spec in Red Hat
catalog.

JIRA: ISV-1857


haripate (approved):
LGTM

PR #300, opened 2022-04-13
[ISV-1953] Fix operator index image signing with manifest list retrival

wcheang:
Previously the pipeline was signing manifest digests straight from
IIB builds, when it should be the manifest list digests that's
retrieved from the registry, using on the IIB digests. This change
fixes that.


wcheang:
@amisstea @haripate still finishing up testing, but would you please take a look and see if the approach makes sense?

wcheang:
@haripate @amisstea this is ready now. Please review. I was not able to run the release pipeline all the way, IIB build failed, but I don't believe it's related to my changes. I tested it as much as I could by running the get-manifest-list-digests task on its own, then feeding the output to the request-signature task. Other tasks seem unaffected by the addition of the new podman user.

wcheang:
Sample log from the task run if it helps at all with the review: http://pastebin.test.redhat.com/1044893

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
```diff
@@ -0,0 +1,77 @@
...
+        # no-op for dev or qa since IIB doesn't run
+        # output dummy/test values for signing purposes
+        if [[ $(params.environment) == "dev" || $(params.environment) == "qa" ]]; then
+          echo -n "registry.redhat.io/redhat/test-operator-index:v4.9" | tee "$(results.docker_references.path)"
+          echo "$(params.manifest_digests)" | awk -F '@' '{print $2}' | tee "$(results.manifest_list_digests.path)"
+          echo "Getting manifest list digests is a NOOP for dev and qa environments at this time."
+          exit 0
+        fi
+
+        DIGEST_LIST=$(echo $(params.manifest_digests) | tr "," " ")
```
haripate, line 39: Doesn't it make sense to put this line of code below line no 47?

haripate: Ah NVMD, I figured out why it is there.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
```diff
@@ -0,0 +1,77 @@
...
+
+        if [[ $(params.environment) != "prod" ]]; then
+          export HTTP_PROXY="http://squid.corp.redhat.com:3128"
+          export HTTPS_PROXY="http://squid.corp.redhat.com:3128"
+
+          # TODO find a better way to set registry based on env
+          DIGEST_LIST=${DIGEST_LIST//registry.redhat.io/registry.stage.redhat.io}
```
haripate, line 46: I think this line of code might be look like this.
```suggestion
          DIGEST_LIST="${DIGEST_LIST}//registry.redhat.io/registry.stage.redhat.io"
```

haripate (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
wcheang: No this is a string replacement: https://www.educba.com/bash-replace-string/

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
```diff
@@ -0,0 +1,77 @@
...
+      description: Docker config for retrieving the bundle image
+  steps:
+    - name: podman-manifest-inspect
+      image: "$(params.pipeline_image)"
+      script: |
+        set -xe
+
+        # no-op for dev or qa since IIB doesn't run
+        # output dummy/test values for signing purposes
+        if [[ $(params.environment) == "dev" || $(params.environment) == "qa" ]]; then
```
haripate, line 32: What if we declare `environment` params as an env variable?

haripate, line 32: What if we declare `environment` params as an env variable?

haripate (changes requested):

haripate (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
haripate: Ah, I see.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
wcheang: That should be easy to do, but why? The only benefit I see is that it'll be slightly shorter.

haripate: Because we want to set the registry based on env in another way in the future, then it will be easy if we are using the env variable rather than referencing the params every time. WDYT?

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
```diff
@@ -0,0 +1,77 @@
...
+              DOCKER_REFERENCES+=$REFERENCE,
+            done
+
+            # parse json output of manifest inspect into comma separated list of manifest digests for signing
+            MANIFEST_LIST_DIGESTS+=$(echo $MANIFEST_LIST | tr " " ",")
+            MANIFEST_LIST_DIGESTS+=","
+
+        done
+
+        echo $DOCKER_REFERENCES | tee "$(results.docker_references.path)"
```
amisstea, line 76: Should some of these `echo` commands use `-n` to strip possible newlines from results?

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -423,68 +423,85 @@ spec:
         - name: kerberos_keytab_secret_key
           value: "$(params.kerberos_keytab_secret_key)"
 
+    # use manifest digests from IIB output and get the manifest list digests
```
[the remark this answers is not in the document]
wcheang: Yeah I think you're right. It would make more sense to swap the names. I will make the changes and re-test.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-manifest-list-digests.yml
wcheang: Updated.

wcheang:
@haripate @amisstea please review again. I've made the suggested changes, and was able to confirm that signature verification works on stage with the new changes.

Allda (approved):

amisstea (approved):

PR #307, opened 2022-05-03
[ISV-1880] Write new value for build_categories

kosciCZ:
SSIA


kosciCZ:
@Allda Could you please take a look? Should be a simple change and it'd be great if it could go to stage today.

Allda (approved):

PR #312, opened 2022-05-12
Remove IBM replication task

Allda:
The task is no longer needed because IBM decided to drop this feature.

JIRA: ISV-2040


jpaulovic:
@acornett21 @jomkz @bcrochet @tonytcampbell FYI

kosciCZ (approved):
It's been a while since I worked on OP, but from what I can tell, the step has been removed and relevant files also. LGTM :+1:

PR #317, opened 2022-06-24
Hardcoded images as parameters

mvalarh:
Existing parameters
- gitInitImage in commit-pinned-digests.yaml

New parameters
- skopeo_image
- digest_pinning_tool_image
- podman_image
- ubi8_minimal_image
- fedora_image
- go_toolset_image
- preflight_trigger_image
- ose_cli_image
- pipelines_cli_tkn_image

Signed-off-by: Martin Vala <mavala@redhat.com>

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
@@ -23,6 +23,9 @@ spec:
         Binding a Secret to this Workspace is strongly recommended over other
         volume types.
   params:
+    - name: gitInitImage
+      description: Git init image
+      default: "registry.redhat.io/openshift-pipelines/pipelines-git-init-rhel8@sha256:bc551c776fb3d0fcc6cfd6d8dc9f0030de012cb9516fac42b1da75e6771001d9"
     - name: dirty_flag
     - name: git_branch
     - name: verbose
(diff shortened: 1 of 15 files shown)


amisstea (approved):

mvalarh:
@amisstea, I would also recommend to put default images in to `pipeline` files? So it will be clear from pipeline configuration what are defaults. Is it ok if i would do it in next step?

ansible/roles/operator-pipeline/templates/openshift/tasks/commit-pinned-digest.yml
```diff
@@ -23,6 +23,9 @@ spec:
         Binding a Secret to this Workspace is strongly recommended over other
         volume types.
   params:
+    - name: gitInitImage
```
Allda, line 26: All other params use snake case formatting, but this one uses camel case. Can you please alight this param with others please?

Allda: Oh... I see it is already merged... nvm

mvalarh: There was `pipeline_image` already so i followed this. I think that `gitInitImage` is convention from `git-clone.yml` and i am guessing that this file was copied from `https://github.com/tektoncd/catalog/blob/main/task/git-clone/0.1/git-clone.yaml` so i left conventions from there

PR #319, opened 2022-06-28
fix missing ( in variable name for image in upload-pipeline-logs.yml

acornett21:
- Fixing a bug that was introduced with image refactor to `default`

Signed-off-by: Adam D. Cornett <adc@redhat.com>

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/upload-pipeline-logs.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/upload-pipeline-logs.yml
@@ -53,7 +53,7 @@ spec:
         optional: true
   steps:
     - name: get-pipeline-logs
-      image: "$params.pipelines_cli_tkn_image)"
+      image: "$(params.pipelines_cli_tkn_image)"
       script: |
         #! /usr/bin/env bash
         set -xe


amisstea (approved):

acornett21:
> Probably because it's still valid YAML. Tekton will also treat this like a standard string rather than perform the variable substitution.

@amisstea Based on the error below it looks like its trying to do the variable substation and failing, or am I miss understanding what you mean?

```
Warning  InspectFailed       14s (x8 over 80s)  kubelet          Failed to apply default image tag "$params.pipelines_cli_tkn_image)": couldn't parse image reference "$params.pipelines_cli_tkn_image)": invalid reference format
  Warning  Failed          14s (x8 over 80s)  kubelet          Error: InvalidImageName
  ```

**Comment by amisstea** [comment]
> > Probably because it's still valid YAML. Tekton will also treat this like a standard string rather than perform the variable substitution.
> 
> @amisstea Based on the error below it looks like its trying to do the variable substation and failing, or am I miss understanding what you mean?
> 
> ```
> Warning  InspectFailed       14s (x8 over 80s)  kubelet          Failed to apply default image tag "$params.pipelines_cli_tkn_image)": couldn't parse image reference "$params.pipelines_cli_tkn_image)": invalid reference format
>   Warning  Failed          14s (x8 over 80s)  kubelet          Error: InvalidImageName
> ```
Is that error originating from a running pipeline? It appears kubelet doesn't like the image ref `$params.pipelines_cli_tkn_image)` which is understandable and indicates no variable substitution was performed.

As for linting, we have a basic YAML validity check in the GH `validate` workflow and also rely on OCP to validate these resources when applied. Neither of them are going to catch a typo like this because it's going to manifest at runtime. Our `E2E-CI` workflow might not pass as a result but it's been failing lately due to unrelated issues :disappointed:.

File: ansible/roles/operator-pipeline/templates/openshift/tasks/upload-pipeline-logs.yml [hunk 1]
```diff
@@ -53,7 +53,7 @@ spec:
         optional: true
   steps:
     - name: get-pipeline-logs
-      image: "$params.pipelines_cli_tkn_image)"
+      image: "$(params.pipelines_cli_tkn_image)"
```
mvalarh, line 56: Thanks for fixing it

PR #326, opened 2022-07-14
Update nginx image version

wcheang:


wcheang:
@amisstea @Allda updated dev nginx-proxy with the ubi9 image and the webhook works again. Please review.

amisstea (approved):
LGTM :+1: 

Do we need to coordinate the release of this change with the prod cluster upgrade or will this also work with the current version?

Allda (approved):

wcheang:
Testing the new image on prod, and the webhook still works fine. I have reverted the change and will do proper deployments next week.

PR #328, opened 2022-07-20
Add support for undistributed operator bundle projects

amisstea:
This requires a Tekton feature-flag modification to change the behavior of
when expressions. The default behavior will switch over in an upcoming
Tekton release.

See https://github.com/tektoncd/pipeline/discussions/4185 for the announcment.

Jira: ISV-2334

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
@@ -419,6 +419,11 @@ spec:
       runAfter:
         - create-container-image
         - get-supported-versions
+      when: &whenNotUndistributed
+        - input: &operatorDistribution "$(tasks.get-cert-project-related-data.results.operator_distribution)"
+          operator: notin
+          values:
+            - undistributed
       taskRef:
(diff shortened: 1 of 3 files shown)


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -443,6 +448,11 @@ spec:
     - name: get-manifest-digests
       runAfter:
         - publish-bundle
+      when:
+        - input: *operatorDistribution
+          operator: notin
+          values:
+            - undistributed
```
mporrato, line 455: Just a suggestion: since we are reusing the same `when` clause all over the place, perhaps it would make sense to create an anchor for the entire input/operator/values block. WDYT?

Allda (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
amisstea: Good idea. Done.

mporrato (approved):

PR #331, opened 2022-08-09
Add index signature verification pipeline

amisstea:
Jira: ISV-2290

Included is a generic PipelineRun listener which sends messages to a
Google Chat upon completion. Events are grouped into threads using the
combination of namespace and event type. This allows chat users to
subscribe to only success or failure events for a particular environment
and purpose such as prod index signature failures or stage operator
pipeline successes.

All cluster-wide configuration is now controlled by Ansible playbooks
because of the need to apply a CloudEvent sink for all Tekton events.
Amongst this configuration is an updated PipelineRun/TaskRun pruning
policy which removes the need to configure a CronJob for this purpose
in each namespace.


Allda (approved):
Looks great!

PR #335, opened 2022-08-24
[ISV-2512] Remove pipelinerun-cleaner CronJob

amisstea:


wcheang:
We should be keeping way more than 5 or 10, but should the job be removed completely? What's the default behavior without the cleanup?

amisstea:
> We should be keeping way more than 5 or 10, but should the job be removed completely? What's the default behavior without the cleanup?

The [tekton pruner](https://tekton.dev/docs/operator/tektonconfig/#pruner) keeps 100 `PipelineRuns` around cluster-wide based on our [config](https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/06514ecf8726e79b9b6b8a49d985e853147f8597/ansible/roles/config-ocp-cluster/files/tektonconfig.yml#L10).

This CronJob, as configured, isn't keeping enough runs around anyway. When there is a flurry of community signing PipelineRuns and some long running hosted or release PIpelineRuns, the 10 run limit in prod is exceeded and this CronJob blindly deletes the oldest. I had to disable this CronJob in prod for this reason.

wcheang:
> The [tekton pruner](https://tekton.dev/docs/operator/tektonconfig/#pruner) keeps 100 `PipelineRuns` around cluster-wide based on our [config](https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/06514ecf8726e79b9b6b8a49d985e853147f8597/ansible/roles/config-ocp-cluster/files/tektonconfig.yml#L10).
Fine with me as long as we don't run out of storage. LGTM :+1:

wcheang (approved):

PR #337, opened 2022-09-01
[ISV-2499] The CI pipeline doesn't support OCP 4.11

mporrato:
- Workaround a Tekton bug affecting the Openshift Pipeline Operator 1.8 where a task output containing valid json is not propagated correctly
- Ensure that task steps involving podman or buildah run as non-root to support the new SCCs in OCP 4.11

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
@@ -47,6 +47,7 @@ spec:
         chmod 755 $DOCKER_CONFIG
         cp $HOME/.docker/config.json $DOCKER_CONFIG/config.json
         chmod 644 $DOCKER_CONFIG/config.json
+        chmod go+rx .
 
     - name: inspect-base-index
       image: "$(params.podman_image)"
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
@@ -33,19 +33,29 @@ spec:
       - name: FBC_FILE
         value: fbc.txt
   steps:
-    - name: inspect-base-index
+    - name: prepare
       image: "$(params.podman_image)"
       workingDir: $(workspaces.output.path)
       script: |
         #! /usr/bin/env bash
-        set -exo pipefail
+        # This is needed to ensure the inspect-base-index step, which runs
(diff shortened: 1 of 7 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
```diff
@@ -63,6 +73,7 @@ spec:
         #! /usr/bin/env bash
         set -x
 
+        chmod 644 podman-inspect.json
```
amisstea, line 76: Why do we need to change the file permissions once more here?

ansible/roles/operator-pipeline/templates/openshift/tasks/get-supported-versions.yml
```diff
@@ -48,11 +48,13 @@ spec:
         echo $VERSION_INFO \
           | jq -r '[.indices[].path]' \
           | tr -d '\n\r' \
+          | base64 \
           | tee $(results.indices.path)
 
         echo $VERSION_INFO \
           | jq -r '[.indices[].ocp_version]' \
           | tr -d '\n\r' \
+          | base64 \
```
amisstea, line 57: Please include a comment about this workaround with any related links to the details. It would also be helpful to mention the base64 encoding in the description of these fields.

amisstea (commented):
Which combinations of OCP and OpenShift Pipelines versions have you tested this against?

ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
mporrato: It's not strictly needed. It's just to avoid leaving world-writable files around.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-supported-versions.yml
mporrato: Yes, that was the plan. I just finished reporting the bug upstream. I'll add a comment with a link to the gh issue.

mporrato:
> Which combinations of OCP and OpenShift Pipelines versions have you tested this against?

I have tesed it against OCP 4.10.25 + Pipelines Operator 1.8, OCP 4.11.0 + Pipelines Operator 1.8, OCP 4.10.25 + latest upstream Tekton Pipelines.

amisstea (approved):
LGTM :+1:

PR #349, opened 2022-09-22
[ISV-2647] Fix how supported index lists are passed

amisstea:
The solves a problem with trailing slashes seen on the index image pull specs in the publish-bundle task. It removes the prior workaround.


wcheang:
So just to make sure, it doesn't hit the tekton bug now even without the base64 encoding?

amisstea:
> So just to make sure, it doesn't hit the tekton bug now even without the base64 encoding?

Correct. The results are no longer valid JSON so there's no need for the base64 encoding workaround.

wcheang (approved):

PR #350, opened 2022-09-28
Move publish-resources to before publish-bundle

wcheang:
The containerRepository needs to be created before the IIB build, since IIB uses registry proxy to resolve the pullspec.


amisstea (approved):

wcheang:
Successful integration test run after the change: https://github.com/redhat-openshift-ecosystem/operator-pipelines/actions/runs/3146471953/jobs/5114952121

PR #356, opened 2022-10-07
[ISV-2398] Implement process to check image signature before release.

wcheang:
This splits up the IIB process of adding bundle images to index and copying the images to the published repo into two separate steps, so that images will not be published before they're signed.

Summary of what's done:
- Update IIB call to not overwrite the `from_index` images. This way IIB will write the images to (unpublished) temporary repos instead of the original published repos. `publish-bundle` is renamed to `add-bundle-to-index` to be more descriptive.
- Update `get-manifest-digests` to get the digests from the temp locations returned by IIB.
- Add a new step `publish-index` after signing to copy the images to the published repos, thus making the new images published.


ansible/roles/operator-pipeline/templates/openshift/tasks/publish-to-index.yml
```diff
@@ -93,20 +71,28 @@ spec:
...
+          echo "Source image: $SRC_IMAGE"
+          VERSION=$(echo $i | awk -F '+' '{print $1}' | awk -F ':' '{print $2}')
+          DEST_IMAGE="${FROM_INDEX}:${VERSION}"
+          echo "Dest image: $DEST_IMAGE"
 
-        echo -n "success" | tee "$(results.status.path)"
-        cat manifest-list-digests.txt | tee "$(results.manifest_list_digests.path)"
+          skopeo \
+            --command-timeout 300s copy \
+            --format v2s2 --all \
```
Allda, line 93: Is th `--format` necessary here? I have seen cases when images were converted from oci to docker schema and digest have been changed. Also, skopeo can fail if this option is used and the input image can't be converted.

Allda (commented):
Changes look good to me but I have a concern. If multiple release pipelines run at the same time, don't they re-write indices if the build and publishing steps are separated?

ansible/roles/operator-pipeline/templates/openshift/tasks/publish-to-index.yml
wcheang: This is what IIB is currently using according to their logs, so I'm just trying to preserve existing behaviors.

wcheang:
@Allda was I able to address all your concerns?

Allda (approved):

PR #357, opened 2022-10-11
DNS level

J0zi:
DNS level


Allda (changes requested):
The CI pipeline can't have variables defined by ansible because we can't force partners to always use ansible. The pipeline is directly used from the repository.

J0zi:
OK, changes reverted. Adding just DNS related updates.

Allda (approved):

mvalarh (approved):

PR #361, opened 2022-10-14
[ISV-2718] Operator release pipeline optionally create latest tag in Quay

mantomas:


ansible/roles/operator-pipeline/templates/openshift/tasks/copy-image.yml
```diff
@@ -10,8 +10,12 @@ spec:
       default: "registry.redhat.io/rhel8/skopeo:8.4-13"
     - name: src_image
       description: reference to the source bundle image
-    - name: dest_image
-      description: reference to the destination bundle image
+    - name: dest_image_first_part
```
Allda, line 13: haha :D let's give it a meaningful name. how about dest_image_registry, dest_image_repository, and dest_image_tag?

ansible/roles/operator-pipeline/templates/openshift/tasks/copy-image.yml
```diff
@@ -23,8 +27,15 @@ spec:
       script: |
         set -xe
 
+        # optionally tag destination image as latest
+        if [ "$(params.is_latest)" = "true" ]; then
```
Allda, line 31: The if/else logic here is wrong. The "optional" means that the image is copied to the destination registry with a given tag and OPTIONALLY the latest tag is also created. This correct logic doesn't support both.

mantomas: logic is changed to reuse `skopeo copy` to retag it `latest` if condition true - original tag is kept in both cases

Allda (approved):

PR #362, opened 2022-10-17
Replace isv guild gpg keys with pipelines keys

Allda:
Pipelines have a custom set of gpg keys. The isv guild keys will be deprecated soon.

JIRA: ISV-2727


Allda:
The PR needs to be merged after secrets are re-encrypted with a new key.

wcheang (approved):

PR #366, opened 2022-10-26
Fix generate-index:prepare to use podman gid 1000

jsm84:
This relates to Jira ISV-2654. PR is linked to EET-3092 in RH Jira.

## Summary
I've discoverd the root cause of the permission denied error when running as the intended uid (1000) in the generate-index: inspect-base-index task step.

The root cause was identified in the preceding step generate-index: prepare where the podman-inspect.json file is given group read/write permissions, effective to gid 0/root.

Typically, the restricted SCC would launch containers as part of the root group on OCP, so it was a safe assumption that group permissions would apply.

There are two issues, however:

1. Podman assumes it can write to the uidmap in the kernel /proc filesystem if it is run as either UID or GID 0.
2. Setting even just the GID to 0 will cause podman to exit due to the uidmap error, requiring the pod to run as privileged. Doing this draws a warning on OCP 4.11, and will likely not work out-of-the-box on OCP 4.12

Therefore, the best course of action was to update the group ownership of podman-inspect.json, as well as other files in the pipeline workspace volume to gid 1000 (podman).

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
@@ -41,12 +41,16 @@ spec:
         # This is needed to ensure the inspect-base-index step, which runs
         # under a different UID, has access to its output file
         truncate -s 0 podman-inspect.json
+        chown :1000 podman-inspect.json
         chmod 664 podman-inspect.json
 
         mkdir -p $DOCKER_CONFIG
+        chown :1000 $DOCKER_CONFIG
         chmod 775 $DOCKER_CONFIG
         cp $HOME/.docker/config.json $DOCKER_CONFIG/config.json
+        chown :1000 $DOCKER_CONFIG/config.json
         chmod 664 $DOCKER_CONFIG/config.json
+        chown :1000 .
         chmod go+rx .
 
     - name: inspect-base-index


bovem:
Thanks @jsm84 . I tested these changes on my environment and the issue was resolved.

My Environment:
```
CRC version: 2.8.0+217b3bd
OpenShift version: 4.11.1
Podman version: 4.1.1

$ tkn version
Client version: 0.25.0
Pipeline version: v0.37.4
Triggers version: v0.20.2
Operator version: v0.60.1

$ oc version
Client Version: 4.11.1
Kustomize Version: v4.5.4
Server Version: 4.11.1
Kubernetes Version: v1.24.0+4f0dd4d

$ podman version
Version:      3.0.2-dev
API Version:  3.0.0
Go Version:   go1.15.7
Built:        Wed Apr  7 09:07:54 2021
OS/Arch:      linux/amd64
```

acornett21 (approved):

yogev-lb:
hey [jsm84](https://github.com/jsm84)

I'm trying to certify a CSI driver on OpenShift, working against CRC cluster:

```bash
CRC version: 2.6.0+d606e64
OpenShift version: 4.10.22
Podman version: 4.1.0
```

Since this is the newest cluster version which runs `OpenShift version: 4.10` and the docs say to target 4.10 at most since 4.11 is not yet supported.

I hit the same error:
`tee: podman-inspect.json: Permission denied`
which your changes may have fixed, but i see that they got reverted by #372 

Is there any fix planned?
Is this ticket still open and being worked on?
What do you recommend I should do?

freeznet:
hitting the same issue here 
```
CRC version: 2.10.1+426703d
OpenShift version: 4.11.7
Podman version: 4.2.0
```

PR #385, opened 2022-12-08
Remove repo published argument in create_container_image.py

johnbieren:
Signed-off-by: Johnny Bieren <jbieren@redhat.com>

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/create-container-image.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/create-container-image.yml
@@ -88,7 +88,6 @@ spec:
           --pyxis-url $PYXIS_URL \
           --connect-registry $REGISTRY \
           --isv-pid $ISV_PID \
-          --repo-published "true" \
           --repository $REPOSITORY \
           --certified "true" \
           --docker-image-digest $CONTAINER_DIGEST \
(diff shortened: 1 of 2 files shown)


johnbieren:
> Can you please provide more details about the change and explain why it is needed?
> 
> I see the argument was defined but never used. Since the pipeline will always produce images with the published set to True I am leaning toward removing the argument completely.
> 
> Also, the argument is stored as a string which is not compatible with the container image schema that requires a boolean.

It isn't needed. I was just trying to fix what looks like a bug (the argument being defined, but not used). I can change the PR to just remove the argument all together then

Allda:
Thank you for the explanation. Since the argument is irrelevant I would instead remove it entirely from a code. If you are ok with removing it please update the PR. Otherwise, we will address this code change in other PR in upcoming releases.

Allda:
@johnbieren Looks good, there is one more place where this arg is used. Please also remove it from there: https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/create-container-image.yml#L91

johnbieren:
> @johnbieren Looks good, there is one more place where this arg is used. Please also remove it from there: https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/create-container-image.yml#L91

fixed, thanks

Allda (approved):

PR #389, opened 2022-12-20
[ISV-2745] [ISV-2746] Update operator-release pipeline to use tekton bundles

haripate:


ansible/roles/operator-pipeline/files/operator-pipeline-config.yml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: operator-pipeline-config
```
kosciCZ, line 4: This does not seem to be the right name for this kind of config, see [here](https://github.com/tektoncd/pipeline/blob/main/docs/install.md#customizing-the-pipelines-controller-behavior). Also, I'm not sure this actually gets applied anywhere.

kosciCZ (changes requested):
There are also same tasks in use in `community-signing-pipeline`, please make sure to replace all occurences.

ansible/roles/config-ocp-cluster/tasks/pipelinerun-listener.yml
```diff
@@ -164,3 +164,18 @@
             name: config-defaults
           data:
             default-cloud-events-sink: "http://el-pipelinerun-listener.{{ pipelinerun_listener_namespace }}.svc.cluster.local:8080"
+
+    - name: Set default Tekton CloudEvents for Tekton Bundles
```
Allda, line 168: The file already exists on the cluster with a bunch more settings. By adding it here you will override the cluster settings. Can you just flip those switches manually instead of overriding the whole file?

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-signing-pipeline.yml
```diff
@@ -40,6 +40,7 @@ spec:
     - name: verify-whitelisted-index-image
       taskref:
         name: verify-whitelisted-index-image
+        bundle: "quay.io/redhat-isv/tkn-signing-bundle:3638365517"
```
Allda, line 43: There is already a newer version of the bundle `3828559997`. Please update it here and below as well.

ansible/roles/config-ocp-cluster/tasks/pipelinerun-listener.yml
```diff
@@ -164,3 +164,4 @@
             name: config-defaults
           data:
             default-cloud-events-sink: "http://el-pipelinerun-listener.{{ pipelinerun_listener_namespace }}.svc.cluster.local:8080"
+
```
Allda, line 167: Can you please remove the extra line?

Allda (approved):

ansible/roles/config-ocp-cluster/tasks/pipelinerun-listener.yml
haripate: Updated!

mvalarh:
@haripate you can try to add
```
- name: "Apply tekton resolvers-feature-flags"
  kubernetes.core.k8s:
    state: present
    definition:
      apiVersion: v1
      kind: ConfigMap
      metadata:
        name: resolvers-feature-flags
        namespace: tekton-pipelines-resolvers
        labels:
          app.kubernetes.io/component: resolvers
          app.kubernetes.io/instance: default
          app.kubernetes.io/part-of: tekton-pipelines
      data:
        # Setting this flag to "true" enables remote resolution of Tekton OCI bundles.
        enable-bundles-resolver: "true"
        # Setting this flag to "true" enables remote resolution of tasks and pipelines via the Tekton Hub.
        enable-hub-resolver: "true"
        # Setting this flag to "true" enables remote resolution of tasks and pipelines from Git repositories.
        enable-git-resolver: "true"
        # Setting this flag to "true" enables remote resolution of tasks and pipelines from other namespaces within the cluster.
        enable-cluster-resolver: "true"
```

mvalarh (approved):

kosciCZ (approved):

PR #392, opened 2023-01-11
Run generate-index as gid 1000

### LINKED ISSUE #387: CI-Pipeline permission denied failure on Azure and CRC
This issue is explained in RH Jira ISV-2654, which I'd already tried resolving unsuccessfully in #366.

The ci-pipeline fails during the `generate-index : inspect-base-index` step when run on both CRC and Azure, with the following error:

```
+ tee -a podman-inspect.json
+ podman image inspect image-registry.openshift-image-registry.svc:5000/rvtest/certified-operator-index:v4.11
tee: podman-inspect.json: Permission denied
```

The root cause is that `podman-inspect.json` has `gid 0` (root group) ownership, but the `inspect-base-index` pod does not run with `gid 0` on either CRC or Azure. When `uid/gid 1000` attempts to write to this file, it fails as expected due to lacking permissions.

jsm84:
Fixes #387

This should work across all known OpenShift architectures, storage plugins, and filesystems (without extended attributes).

Signed-off-by: Josh Manning <19478595+jsm84@users.noreply.github.com>

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
@@ -41,6 +41,7 @@ spec:
         # This is needed to ensure the inspect-base-index step, which runs
         # under a different UID, has access to its output file
         truncate -s 0 podman-inspect.json
+        chown :1000 podman-inspect.json
         chmod 664 podman-inspect.json
 
         mkdir -p $DOCKER_CONFIG
@@ -54,6 +55,7 @@ spec:
       workingDir: $(workspaces.output.path)
       securityContext:
         runAsUser: 1000
+        runAsGroup: 1000
       script: |
         #! /usr/bin/env bash
         set -exo pipefail


jsm84:
@mayurwaghmode Please test this on your environment.

mayurwaghmode:
> @mayurwaghmode Please test this on your environment.

@jsm84 I have tested this on ppc64le's OCP 4.10 and 4.11, and it is working as expected. Thank you so much

bovem:
Thanks @jsm84 . I was facing the permissions issue on OCP 4.12-nightly cluster and this fix resolved it.

Allda:
@mporrato Can you please check if this change is compatible with the permission issues you fixed this week? If so we can merge it.

mporrato:
> @mporrato Can you please check if this change is compatible with the permission issues you fixed this week? If so we can merge it.

Just ran an integration test with this patch applied on top of the current main branch and it seems to work just fine, but I can't test on CRC right now.

PR #401, opened 2023-01-20
[ISV-2659] operator_name check in hosted pipeline

mantomas:


operator-pipeline-images/operatorcert/entrypoints/check_operator_name.py
```diff
@@ -0,0 +1,79 @@
...
+
+def setup_argparser() -> argparse.ArgumentParser:  # pragma: no cover
+    """
+    Setup argument parser
+
+    Returns:
+        Any: Initialized argument parser
+    """
+    parser = argparse.ArgumentParser(
+        description="Reserve the given operator package name"
```
wcheang, line 20:
```suggestion
        description="Check if given association has any operator name already reserved"
```

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
```diff
@@ -226,6 +227,29 @@ spec:
           workspace: pipeline
           subPath: src
 
+    - name: get-cert-project-related-data
```
wcheang, line 230: Is it actually okay to just add this task to the CI pipeline? It doesn't just get cert project data, it also performs validations, and if those validations are not met, the task fails. This means that we're introducing requirements to the CI pipeline that weren't there before.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
```diff
@@ -273,9 +297,31 @@ spec:
           workspace: pipeline
           subPath: src
 
-    - name: dockerfile-creation
+    # check if cert_project already has package_lock
+    # and if the package_name match operator_name
+    - name: check-operator-name
```
wcheang, line 302: I know adding this to the CI pipeline was requested in the jira, but I wonder if it's actually necessary when the check is already in the hosted pipeline? Granted it'll help partners detect this specific problem early, but we don't duplicate all hosted pipeline validations in the CI pipeline. Maybe @Allda has a better idea on what the criteria on adding validations in the CI pipeline are.

acornett21: I don't think we should add this check to the CI pipeline, since we would now force the partner to run with a `full` set of tokens/keys to talk to Red Hat. If we put this check in we need to change all of our documentation (GitHub and Policy docs) to no longer allow a minimal run and are taking away the iterative approach that partners use the CI pipeline for.

ansible/roles/operator-pipeline/templates/openshift/tasks/check-operator-name.yml
```diff
@@ -0,0 +1,66 @@
...
+kind: Task
+metadata:
+  name: check-operator-name
+spec:
+  params:
+    - name: pipeline_image
+    - name: association
+      description: The association that the operator belongs in, usually the isv_pid.
+    - name: operator_name
+      description: The operator name to be reserved.
```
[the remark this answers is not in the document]
mantomas: file removed

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
mantomas: decided not to edit CI pipeline

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
mantomas: Decided not to add this check to CI pipeline - all changes are now only in hosted pipeline

operator-pipeline-images/operatorcert/entrypoints/reserve_operator_name.py
```diff
@@ -32,6 +32,38 @@ def setup_argparser() -> argparse.ArgumentParser:  # pragma: no cover
...
+
+    rsp.raise_for_status()
+
+    packages = rsp.json().get("data")
+
+    if packages:
+        # there should only be 1 package for given association/isv_pid
+        package = packages[0]
+        if package["package_name"] != args.operator_name:
+            LOGGER.error(
```
Allda, line 51: These 2 messages will be directly visible to users. Can you reword them a little so the partner understands what's wrong? I mean use wording that is not API specific and is understandable even though the user doesn't know anything about API details.

operator-pipeline-images/operatorcert/entrypoints/reserve_operator_name.py
```diff
@@ -32,6 +32,38 @@ def setup_argparser() -> argparse.ArgumentParser:  # pragma: no cover
     return parser
 
 
+def check_operator_name_registered_for_association(args) -> None:
```
Allda, line 35: Please add a docstring.

mantomas: added

operator-pipeline-images/operatorcert/entrypoints/reserve_operator_name.py
mantomas: Messages updated

Allda (approved):

wcheang (approved):

PR #411, opened 2023-02-07
ISV-3212 Signature Verification Pipeline fails after OSD PROD upgrade

mgreczi:
ISV-3212


Allda:
@mgreczi Can you please provide a summary of this change and how exactly it helps with the issue?

mporrato:
@Allda the problem is that the `podman image trust set` command tries to update `/etc/containers/policy.json` but UID 1000 cannot change that file.
The idea was to pre-populate policy.json with the correct public key and drop the image trust command entirely.

ansible/roles/index-signature-verification/files/tasks/verify-index-signatures.yml
```diff
@@ -36,6 +39,9 @@ spec:
     - name: keys
       configMap:
         name: "$(params.keys_config_map)"
+    - name: policies
```
Allda, line 42: How does it know where to mount the volume? Other volumes are mounted to a specific location, but this one isn't.

mgreczi: Signature verification now works. I update this Pull Request with the working changes.

mporrato (approved):

Allda (approved):

PR #413, opened 2023-02-09
[ISV-1207] Use JWT instead of basic auth for Hydra API

cit1zen:
Basic auth is being decomissioned, which means that we have to switch to OICD/JWT authentication.


cit1zen:
I tested a pipeline that failed on the HYDRA step, and it passed. You can find it here: https://console-openshift-console.apps.pipelines-stage.0ce8.p1.openshiftapps.com/k8s/ns/aormandy-test/tekton.dev~v1beta1~PipelineRun/operator-hosted-pipeline-run-v88wh

ansible/roles/operator-pipeline/tasks/pipeline-secrets.yml
```diff
@@ -272,8 +272,9 @@
           suffix: "{{ suffix }}"
           env: "{{ env }}"
       data:
-        username: "{{ lookup('file', operator_pipeline_hydra_username_local_path, rstrip=False) | b64encode }}"
-        password: "{{ lookup('file', operator_pipeline_hydra_password_local_path, rstrip=False) | b64encode }}"
+        sso_token_url: "{{ operator_pipeline_hydra_sso_token_url | b64encode }}"
```
wcheang, line 275: The SSO token url is just plain text in the config and not really that private. I don't think it's necessary to save it inside a secret just to extract it back out later?

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -51,12 +51,17 @@ spec:
...
+      description: |
+        The key within the Kubernetes secret that contains Hydra SSO client ID.
+      default: sso_client_id
+    - name: hydra_secret_sso_client_secret_key
+      description: |
+        The key within the Kubernetes secret that contains the Hydra SSO password.
+      default: sso_client_secret
+    - name: hydra_secret_sso_token_url_key
+      description: The key within the Kubernetes secret that contains SSO URL.
+      default: sso_token_url
```
wcheang, line 64:
```suggestion
    - name: hydra_secret_sso_token_url
      description: The Red Hat SSO URL.
      default: https://auth.redhat.com/auth/realms/EmployeeIDP/protocol/openid-connect/token
```
If the sso_token_url is not saved within a secret (like I suggested above), then this will need to be changed as well, and tasks that use this var will need to be adjusted accordingly.

Allda (approved):
I agree with @wcheang's suggestion, but otherwise, it looks good.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
Allda: Also if you do the change you need to change also pipeline triggers as they pass non-default parameters.

Allda (approved):
Looks fine to me now!

PR #423, opened 2023-02-28
Create custom SCC with privilege escalation

Allda:
The OpenShift pipelines 1.9 no longer allow running containers with privilege escalation. Since a subset of pipeline tasks requires the privilege escalation we need to use a custom SCC that is build on top of pipeline SCC and add the escalation.

This commit adds the custom SCC, links it with the pipeline service account and refactors ansible to be compatible with the new setup.

JIRA: ISV-3280

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
@@ -38,27 +38,21 @@ spec:
       workingDir: $(workspaces.output.path)
       script: |
         #! /usr/bin/env bash
-        # This is needed to ensure the inspect-base-index step, which runs
-        # under a different UID, has access to its output file
-        truncate -s 0 podman-inspect.json
-        chown :1000 podman-inspect.json
         chmod 664 podman-inspect.json
 
         mkdir -p $DOCKER_CONFIG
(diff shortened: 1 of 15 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
```diff
@@ -55,10 +55,9 @@ spec:
       env:
         - name: STORAGE_DRIVER
           value: vfs
-      workingDir: $(workspaces.output.path)
       securityContext:
-        runAsUser: 1000
-        runAsGroup: 1000
+        privileged: true
+      workingDir: $(workspaces.output.path)
```
mporrato, line 60: Now that this is running as root, we can probably get rid of most of the permission tweaks in the previous step and replace the `tee -a` in this one with a regular stdout redirection.
The chmod in the `determine-index-format` step can also be dropped, I think.

mporrato (approved):

ansible/roles/operator-pipeline/templates/openshift/tasks/generate-index.yml
Allda: Fixed and tested on stage cluster.

PR #426, opened 2023-03-07
Separate role binding for each namespace

Allda:
The role-binding needs to be created with unique name given by namespaces. The original hardcoded name caused that different environments always overwrite the bindings and we were loosing roles after each re-deployment.

JIRA: ISV-3396

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/openshift-pipeline-sa-scc-role-bindings.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/openshift-pipeline-sa-scc-role-bindings.yml
@@ -2,7 +2,7 @@
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: "system:openshift:scc:pipelines-custom-scc"
+  name: "pipeline-{{ oc_namespace }}-pipelines-custom-scc"
 subjects:
   - kind: ServiceAccount
     name: pipeline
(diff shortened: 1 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/openshift-pipeline-sa-scc-role-bindings.yml
```diff
@@ -2,7 +2,7 @@
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: "system:openshift:scc:pipelines-custom-scc"
+  name: "pipeline:pipelines-custom-scc:{{ oc_namespace }}"
```
mporrato, line 5: Can we use a more consistent naming scheme? If we want to match the role binding for the admin service account, this would become something like:
```suggestion
  name: "pipeline-{{ oc_namespace }}-pipelines-custom-scc"
```

mporrato (approved):
Just a small cosmetic suggestion. Otherwise, LGTM.

ansible/roles/operator-pipeline/templates/openshift/openshift-pipeline-sa-scc-role-bindings.yml
wcheang: To add to that, IIRC the oc namespaces are formatted "operator-pipeline-<env>", so there'll be 3 different mentions of"pipeline" all close together. Is that a bit much? Maybe just something like "{{ oc_namespace }}-custom-scc"?

Allda: yep, changed

PR #429, opened 2023-03-16
[ISV-3366] Ignore deleted package lock objects

Allda:
When searching for existing package lock ignore object with deleted flag set to true. The deleted flag was recently added to be able reuse package multiple times.


Allda:
This should be merged after ISV-3365 is implemented.

PR #431, opened 2023-03-21
[ISV-3340] Disable the concurrent pipeline runs for operator release pipeline.

haripate:


ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -0,0 +1,69 @@
...
+  params:
+  - name: lease-name
+    type: string
+  - name: timeout
+    type: string
+    default: 90m
+  stepTemplate:
+    env:
+      - name: TOTAL_DURATION_IN_SECONDS
+        value: ${0}
```
mporrato, line 15: What's the purpose of this? I don't think it's needed.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -547,6 +556,16 @@ spec:
 
 
   finally:
+    - name: release-lease
+      when:
+        - input: $(tasks.status)
+          operator: in
+          values: ["Succeeded", "Completed"]
```
Allda, line 563: What's the point of this condition? Shouldn't the lock be released no matter what's the status?

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: acquire-lease
+spec:
+  params:
```
Allda, line 6: Can you please add a description to the pipeline and to all arguments?

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -0,0 +1,69 @@
...
+  - name: timeout
+    type: string
+    default: 90m
+  stepTemplate:
+    env:
+      - name: TOTAL_DURATION_IN_SECONDS
+        value: ${0}
+  steps:
+    - name: create-lease
+      image: image-registry.openshift-image-registry.svc:5000/openshift/cli:latest
```
Allda, line 18: Is this pullspec always available in the cluster? Can you check a Red Hat Catalog and find the same image in the official Red Hat registry?

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -0,0 +1,69 @@
...
+          echo "  leaseDurationSeconds: ${TOTAL_DURATION_IN_SECONDS}" >> e2e-lease.yaml
+          echo "  holderIdentity: $(params.lease-name)" >> e2e-lease.yaml
+        }
+        # Calculate duration in seconds to store in lease
+        calculate_duration_in_seconds $(params.timeout)
+        # Create yaml that will be used for the lease object
+        create_lease_yaml
+        # Acquire Lease
+        export -f acquire_lease
+        timeout $(params.timeout) bash -c "acquire_lease"
```
Allda, line 69: Missing new line at the end of the file. Please add new line here.

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
haripate: The TOTAL_DURATION_IN_SECONDS variable is needed to be an env variable because it wasn't straight forward to use the return value from one defined function to another defined function. After declaring the variable as an env variable, that issue got resolved.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
haripate: I think yes we don't need this when condition here. I will remove this.

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
haripate: I will check for the image from the Red Hat Catalog and update it here.

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
mporrato: You are not declaring TOTAL_DURATION_IN_SECONDS as an env var here: you are telling tekton to run the script with TOTAL_DURATION_IN_SECONDS set to something undefined (either the literal string `${0}` or it will get resolved to the name of the shell being executed), so I don't think it makes a lot of sense as it is.
Since `calculate_duration_in_seconds()` is the first function you call, it will set the value of `TOTAL_DURATION_IN_SECONDS` for the rest of the script.
What you are missing is an `export TOTAL_DURATION_IN_SECONDS` at the end of `calculate_duration_in_seconds()` because you are executing `acquire_lease()` in a subshell.

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -0,0 +1,69 @@
...
+              echo "Waiting for lease..."
+              acquiredTime=$(oc get lease $(params.lease-name) -o=jsonpath='{.spec.acquireTime}')
+              acquiredTimeInSeconds=$(date +%s --date="${acquiredTime}")
+              currentTimeinSeconds=$(date +%s)
+              elapsedSeconds=$(($currentTimeinSeconds-$acquiredTimeInSeconds))
+              if (( elapsedSeconds > TOTAL_DURATION_IN_SECONDS)); then
+                echo "Lease abandoned, deleting"
+                kubectl delete lease $(params.lease-name)
+              fi
+            done
```
mporrato, line 48: If the lease is being held by another pipelinerun, this will keep hammering the openshift api as fast as it can.
I would add a sleep of a few seconds in the loop.

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -31,24 +27,44 @@ spec:
...
+                echo "TOTAL_DURATION_IN_SECONDS: ${TOTAL_DURATION_IN_SECONDS}"
+                echo "elapsedSeconds: ${elapsedSeconds}"
+                if (( elapsedSeconds > TOTAL_DURATION_IN_SECONDS)); then
+                  echo "Lease abandoned, deleting"
+                  kubectl delete lease $(params.lease-name)
+                  kubectl create -f e2e-lease.yaml
+                  leaseOwner=$(kubectl get lease $(params.lease-name) -o=jsonpath='{.spec.holderIdentity}')
+                  echo "The owner of the lease is ${leaseOwner}"
+                  echo "Acquired lease $(params.lease-name)"
+                  break
```
mporrato, line 61: This can be dropped as the same action will be executed at the start of the next loop.

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -68,4 +84,4 @@ spec:
         create_lease_yaml
         # Acquire Lease
         export -f acquire_lease
-        timeout $(params.timeout) bash -c "acquire_lease"
+        bash -c "acquire_lease"
```
mporrato: Having removed the timeout wrapper, I don't think you need to run `acquire_lease()` in a subshell anymore and as a result, the `export -f acquire_lease` and `export TOTAL_DURATION_IN_SECONDS` can also be dropped.

mporrato (approved):

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - name: lease-name
+      type: string
+      description: The name of the resource which will be acquired.
+    - name: timeout
+      type: string
+      default: 90m
+      description: The timeout for the wait period of leased resource.
+  steps:
+    - name: create-lease
+      image: quay.io/operator_testing/operator-test-playbooks:dev-tekton
```
Allda, line 16: Who owns this repository and what's the support level? Isn't there more trusted sources in the Red Hat registry? I really want to avoid breaking our pipeline in production when somebody decides to remove the image or tag from a repository that we don't own. 

Also please use manifest digests instead of tags.

ansible/roles/operator-pipeline/templates/openshift/tasks/release-lock.yaml
```diff
@@ -0,0 +1,13 @@
...
+metadata:
+  name: release-lease
+spec:
+  params:
+    - name: lease-name
+      type: string
+      description: The name of the resource which will be acquired.
+  steps:
+    - name: delete-lease
+      image: quay.io/operator_testing/operator-test-playbooks:dev-tekton
```
Allda, line 12: Same here.

ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
Allda: This looks like a good official replacement: https://catalog.redhat.com/software/containers/openshift-pipelines/pipelines-cli-tkn-rhel8/60d062834924e14519331697 Please verify if it has all the required packages that you need.

haripate: I have verified that the above red hat registry doesn't have required packages and that is why I figured out to use our own operator-pipeline-images with a bunch of changes in commands for the task steps. Can you please review the changes? 

PS: I have tested all of the changes on the operator release pipeline for my local system and it is working fine.

cc @Allda

Allda: Can you please update the image reference to the one that you suggested in the previous comment? Are you referring to this one? quay.io/redhat-isv/operator-pipelines-images ?

PR #449, opened 2023-05-29
[ISV-2905] fix difference in md5sum in ci vs hosted

mantomas:
Add git cleanup to `verify-pinning` step to avoid different md5sum between ci and hosted pipeline. When  digest-pinning is enabled, `relatedImages` section in csv can be reordered/changed without any actual pinning, causing md5sum difference.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
@@ -85,7 +85,7 @@ spec:
           echo "The branch is reset to its pre-pin state because other unrelated changes may occur during the pin step and need to be undone."
           git checkout -- .
         fi
-    - name: verify-related-imagesmultiline echo
+    - name: verify-related-images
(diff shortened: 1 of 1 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
```diff
@@ -82,8 +82,10 @@ spec:
         else
           echo "Manifests are pinned."
           echo -n "false" | tee $(results.dirty_flag.path)
+          echo "The branch is reset to its pre-pin state because other unrelated changes may occur during the pin step and need to be undone."
+          git checkout -- .
         fi
-    - name: verify-related-images
+    - name: verify-related-imagesmultiline echo
```
Allda, line 88: This looks like an unwanted modification.

mantomas: yes, thanks!

Allda (approved):

PR #451, opened 2023-06-01
Fix ISV-2905 - enable git operations on unowned repository

mantomas:


ansible/roles/operator-pipeline/templates/openshift/tasks/digest-pinning.yml
```diff
@@ -73,6 +73,9 @@ spec:
           exit 0
         fi
 
+        # configure git to enable operations on unowned repository
+        git config --global --add safe.directory "/workspace/source"
```
Allda, line 77: How about `$(workspaces.source.path)` to not hard code the path on multiple places?

mantomas: Thanks, shame on me - will push update.

Allda (approved):

PR #456, opened 2023-06-19
[CLOUDWF-9238] Remove "status" from results in tekton task publish-to-index

bovem:
Signed by : apal@redhat.com


Allda:
@bovem Thanks for the PR. I am surprised this bug passed the integration tests in the first place. Looking at the code again and it seems the status argument was wrongly used in this task and should be removed completely. Could you please edit the PR and remove the line from the task and also remove the `result` section?
`echo -n "success" | tee "$(results.status.path)"`

bovem:
Thanks @Allda I've removed the `status` from the results.

Allda (approved):

PR #458, opened 2023-06-20
[ISV-3647] Verify cert project is not archived

jedinym:


Allda (approved):
Changes look good :+1:

PR #464, opened 2023-07-18
Enhanced regex and err msg for PR title validation

jsm84:
Fixes ISVISSUE-420

Summary:
* The standard regex for PR title validation now checks for a simplified semantic version string (accepting only hyphenated build tags)
* There is a specific check for prefix "v" in the semver portion of a PR title that provides additional error info

Update 1:
* There was one duplicate test case that I changed to validate that a
  prefixed version such as (v1.0.1) would fail.
* There was another test case with a non-semantic version of (1) that is
  no longer valid with this new code

Update 2:
* Correct linting error due to new code in parse_pr_title function
* Elaborate in comment that the new code is for a specific failure case

### DIFF
--- a/operator-pipeline-images/operatorcert/__init__.py
+++ b/operator-pipeline-images/operatorcert/__init__.py
@@ -261,13 +261,21 @@ def parse_pr_title(pr_title: str) -> Tuple[str, str]:
     If yes, extract the Bundle name and version.
     """
     # Verify if PR title follows convention- it should contain the bundle name and version.
-    # Any non- empty string without whitespaces makes a valid version.
-    regex = rf"^operator ([a-zA-Z0-9-]+) \(([^\s]+)\)$"
+    regex = rf"^operator ([a-zA-Z0-9-]+) \(([0-9]+\.[0-9]+\.[0-9]+[a-zA-Z0-9-]*)\)$"
     regex_pattern = re.compile(regex)
+    err_msg = f"Pull request title {pr_title} does not follow the regex 'operator <operator_name> (<version>)"
+
+    # Check <version> for prefix 'v'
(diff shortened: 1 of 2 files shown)


jsm84:
@Allda please review upon your return to office tomorrow.

operator-pipeline-images/operatorcert/__init__.py
```diff
@@ -261,14 +261,20 @@ def parse_pr_title(pr_title: str) -> Tuple[str, str]:
...
+    regex = rf"^operator ([a-zA-Z0-9-]+) \(([0-9]+\.[0-9]+\.[0-9]+[a-zA-Z0-9-]*)\)$"
     regex_pattern = re.compile(regex)
+    err_msg = f"Pull request title {pr_title} does not follow the regex 'operator <operator_name> (<version>)"
+
+    # Check <version> for prefix 'v' failure case and include additional error context
+    regex_v = rf"^operator ([a-zA-Z0-9-]+) \((v[0-9]+\.[0-9]+\.[0-9]+[a-zA-Z0-9-]*)\)$"
+    regex_pattern_v = re.compile(regex_v)
+    err_msg_v = f" where <version> does not contain prefix 'v'"
+
+    if regex_pattern_v.match(pr_title):
```
Allda, line 273: I read this code several times and I have a hard time understanding it. Please correct me if my understanding is wrong but it seems to me that the `v` prefix is not allowed anymore. Is that what you want to achieve here? If so how about operators that passed the certification and have the `v` prefix: https://github.com/redhat-openshift-ecosystem/certified-operators/pull/2523 ?

Allda: Thanks for the explanation @jsm84 . I am ok with merging it. Thanks for your contribution!

Allda (approved):

PR #470, opened 2023-07-31
[ISV-3493] Detect bundles affected by a PR

mporrato:
Sorry for the huge diff but most of it is boilerplate files for the test cases for the operator_repo module.
I've tested it fairly extensively on the preprod repo and seems to work as expected.


operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
```diff
@@ -0,0 +1,245 @@
...
+        # have been affected by the PR
+
+        git_repo_head = GitRepo(repo_path)
+        # Ensure the head commit is checked out (mostly needed for tests)
+        git_repo_head.head.reference = git_repo_head.create_head(
+            "pr_head", head_commit, force=True
+        )
+        git_repo_head.head.reset(index=True, working_tree=True)
+        # Make a temporary clone of the repo and check out the base commit
+        git_repo_base = git_repo_head.clone(repo_clone_path)
```
Allda, line 117: Do we need to clone the repo again if the previous pipeline task already cloned it?

ansible/roles/operator-pipeline/templates/openshift/tasks/parse-repo-changes.yml
```diff
@@ -0,0 +1,78 @@
...
+        jq -r '.deleted_bundles|join(",")' </tmp/changes.json >$(results.deleted_bundles.path)
+
+        function fail() {
+            echo "ERROR: $*"
+            exit 1
+        }
+
+        jq -e '.extra_files|length==0' </tmp/changes.json \
+            || fail "The PR affects non-operator files"
+        jq -e '.affected_operators|length==1' </tmp/changes.json \
```
Allda, line 73: Do you want to support adding multiple operator bundles in the single pipeline? How are we going to handle it in the rest of the pipeline?

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
```diff
@@ -39,6 +40,8 @@ spec:
           value: $(params.git_fork_url)
         - name: revision
           value: $(params.git_pr_branch)
+        - name: depth
+          value: 25
```
Allda, line 44: Where does this number come from?

.yamllint
```diff
@@ -1,6 +1,9 @@
 ---
 extends: default
 
+ignore: |
+  operator-pipeline-images/tests/data/operator-repo
```
Allda, line 5: Do you think we could remove this line and make a repo compliant with a linting?

operator-pipeline-images/tests/test_operator_repo.py
```diff
@@ -0,0 +1,76 @@
...
+    assert "operator-e2e" in repr(op1)
+    assert set(repo) == {op1, op2, op3, op4}
+    assert len(set(op1)) == 1
+    assert len(set(op2)) == 2
+    assert op2 < op1
+    assert op1.root() == repo.root() + "/operators/operator-e2e"
+    assert op1.config == {}
+
+
+def test_bundle() -> None:
```
Allda, line 48: Do you think it would be possible to break this single test into multiple smaller ones so it is easier to maintain it and contribute to the code?

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
mporrato: This is because we need to check out the base commit from the repo and by default the git-clone task makes a shallow clone of depth 1 (from the head commit). 25 should be high enough to include the base commit in almost all cases.

.yamllint
mporrato: I expect to move the operator_repo module, and therefore the test data out of this repo at some point.
I don't think it's worth the effort to try to fix all the yamls.

ansible/roles/operator-pipeline/templates/openshift/tasks/parse-repo-changes.yml
mporrato: Since this is one of the features supported in the current implementation, I'd like to keep it if possible. I think we should be able to handle it with the Matrix construct in tekton. If it will prove to be too difficult, we can always add the relevant check here and abort the pipeline upfront.

Allda: I am a bit worried about this feature because it will complicate all the remaining tasks in the Tekton pipeline and it also makes prow jobs more complicated than needed which increases the complexity that we will need to provide when dealing with issues.

If we want to provide the same user experience as for the ISV pipeline we should keep the logic the same (1PR == 1 operator bundle).

mporrato: Fair enough. I've just added the validation on the number of added bundles.

Allda (approved):

PR #471, opened 2023-08-01
Revert PR title version semver regex

jsm84:
Reverts #464 and #467

The PR title version validation introduced in #464 was meant to resolve ISVISSUE-420 and bring the PR title, bundle directory, and operator version (from the CSV.yaml) into alignment.

It was a simplified regex meant to enforce a rough semantic version in the PR title. This caused issues for another partner, where dots were used in the PR title version's suffix (which, are also semver-compliant). PR #467 was introduced to fix this bug.

The partner is still facing certification issues, specifically failing the PR title validation check of the hosted-pipeline, and this is occurring before errors/results get posted to the partner API.

Thus, the aforementioned PR's should be rolled back until further notice.


jfrancin (approved):
LGTM.

PR #473, opened 2023-08-04
Show pipeline status using PR labels

Allda:
The new Python module has been added to manage labels in Github PRs. The module allows adding and removing labels.

The new task is introduced which sets labels for a PR. The task is used in all pipelines and indicates a status of the pipeline by setting appropriate label.

JIRA: ISV-3895


acornett21:
@Allda If we have to re-run a given pipeline, I assume the tags get `rest/removed`  and go back to `started` -> `pass/fail` correct?

Allda:
> @Allda If we have to re-run a given pipeline, I assume the tags get `rest/removed` and go back to `started` -> `pass/fail` correct?

Yes, that's correct. When you re-trigger a pipeline a labels are cleared and set again based on a status.

ezopezo (approved):

PR #475, opened 2023-08-10
Parse supported ocp version

Allda:
The OCP version range given by the operator is used to query Puxis to get list of supported ocp versions. The task from the ISV pipeline is reused and extended to also support list of versions that should be skipped.

JIRA: ISV-3834


mporrato (approved):
LGTM, but I can see a potential issue when a new ocp version is introduced in pyxis but the correspondent prow job has not been configured yet: the pipeline will fail waiting for the prow tests for the new version to complete. We need to ensure that prow jobs are always created before the pyxis entry is created: the prow job will fail but the pipeline won't be checking it.
Similarly for retirement of old ocp versions, we will need to ensure that the ocp versions are always removed from pyxis first.

PR #477, opened 2023-08-21
[ISV-3495] Create tekton task to perform formal checks on an operator

mporrato:
Implemented checks:

operator:
 - check_operator_name: ensures the operator name in the CSVs is consistent across all operator's bundles

bundle:
 - check_osdk_bundle_validate: runs `operator-sdk bundle validate --select-optional name=operatorhub` against the bundle
 - check_required_fields: checks the presence of the mandatory CSV files as defined in https://github.com/operator-framework/community-operators/blob/master/docs/packaging-required-fields.md#required-fields-for-operatorhub

Because of the complexity of the data structure, the check results are saved to a json file instead of being returned as results by the tekton task.

Moved the operator-repo library to its own repo.


operator-pipeline-images/operatorcert/static_tests/community/validations.py
```diff
@@ -0,0 +1,129 @@
+import binascii
```
Allda, line 1: Can you please add a propper docstrings to the module and all functions?

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -0,0 +1,90 @@
+import json
```
Allda, line 1: Same here, please. It seems we are missing the pylint and mypy in the validation pipeline. I'll file a ticket for that but it will make the followup task easier if we add the docstring to a new code.

Allda (approved):
I have just minor docstring related comments but a logic and new task looks fine to me.

PR #481, opened 2023-08-24
[ISV-3494] check permissions against ci

tomasbakk:


tomasbakk:
i'll go through the files and check the strange changes for the other tasks. seems like it also pushed the things i did not touch but due to some reasons they are changed in this PR

tomasbakk:
looks alright now

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
```diff
@@ -0,0 +1,100 @@
...
+          --git-username="$(params.git_username)" \
+          --operator-name="$(params.operator_name)" \
+          --output-file /tmp/output.json \
+          --verbose
+        list_of_reviewers=$(jq -r '.all_reviewers[]' /tmp/output.json)
+        bool_reviewer="$(jq -r '.is_reviewer' /tmp/output.json)"
+
+        # redirecting results variables to /dev/null so they would not be printed
+        echo -n "$list_of_reviewers" | tee $(results.list_reviewers.path) > /dev/null
+        echo -n "$bool_reviewer" | tee $(results.author_is_reviewer.path) > /dev/null
```
mporrato, line 47: We have the `yq` tool in the container image: we don't need the python script at all to do this:

```sh
CI="operators/$(params.operator_name)/ci.yaml"
if [ -e $CI ] ; then
  yq -r '.reviewers' $CI >$(results.list_reviewers.path)
  yq -r ".reviewers|index(\"$(params.git_username)\") != null" $CI >$(results.author_is_reviewer.path)
else
  echo >$(results.list_reviewers.path)
  echo false >$(results.author_is_reviewer.path)
fi
```

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
```diff
@@ -0,0 +1,100 @@
...
+            for reviewer in $CI_REVIEWERS; do
+              curl -L \
+                -X POST \
+                -H "Accept: application/vnd.github.v3+json" \
+                -H "Authorization: bearer ${GITHUB_TOKEN}" \
+                -H "X-GitHub-Api-Version: 2022-11-28" \
+                -d '{"reviewers": ["'"$reviewer"'"]}' \
+                $API_URL_REVIEWERS > /dev/null
+              echo "Requesting review from reviewers in ci.yaml file: $reviewer"
+            done
```
mporrato, line 98: Way too much curl sorcery for my taste here :mage:. Can we update the dockerfile for the container image to add the `gh` command line tool package (which is available in the standard fedora repo) and use that instead?
It would make our life a lot easier like:

```sh
gh pr edit $(params.pr_url) --add-reviewer `echo $CI_REVIEWERS | xargs | tr ' ' ','`
gh pr review $(params.pr_url) --approve
```

mporrato (changes requested):
We can make this a lot simpler.
Also, don't forget to to change the `runAfter` of the next task in the pipeline.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
tomasbakk: I tried to work with gh package here but it seems that the way how it needs to be authenticated to work with post requests is different from what we already have. I went through the documentation and tried to supply it with token from the env variable as well as calling it with --with-token argument and passing the token this way but I was not able to authenticate the requests. 
But I noticed that we already have github module in operator-pipelines repo https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/operator-pipeline-images/operatorcert/github.py so I can rework the python script to include the post requests there. Then I can get rid of all the curls from bash and also write the tests for the whole thing. WDYT?

tomasbakk: I updated token's scope, tested it and it works now. Thanks for the inputs guys.

tomasbakk:
@mporrato @ezopezo can you please have a look at the recent changes? the script was tested yesterday in my testing PR and it works alright. thanks to Maurizios suggestions i was able to get rid of python script entirely. thank you

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
```diff
@@ -0,0 +1,79 @@
...
+          echo >$(results.list_reviewers.path)
+          echo false >$(results.author_is_reviewer.path)
+        else
+          if grep -q 'reviewers:' $CI; then
+            yq -r '.reviewers[]' $CI >$(results.list_reviewers.path)
+            yq -r ".reviewers|index(\"$(params.git_username)\") != null" $CI >$(results.author_is_reviewer.path)
+          else
+            echo >$(results.list_reviewers.path)
+            echo false >$(results.author_is_reviewer.path)
+          fi
```
mporrato, line 46: Would be more reliable and easier to read this way:
```suggestion
          yq -r '(.reviewers//[]))[]' $CI >$(results.list_reviewers.path)
          yq -r ".reviewers//[]|index(\"$(params.git_username)\") != null" $CI >$(results.author_is_reviewer.path)
```

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
```diff
@@ -0,0 +1,79 @@
...
+          # requesting review from PR author if it is included in the reviewers field
+          if [ "$AUTHOR_IS_REVIEWER" = true ]; then
+            gh pr review $(params.pr_url) --approve
+            echo "This pull request was automatically approved due to PR author is in the reviewers list."
+          else
+            # request a review from the reviewers listed in ci.yml file if the author is not listed in ci.yaml
+            for reviewer in $CI_REVIEWERS; do
+              gh pr edit $(params.pr_url) --add-reviewer `echo $reviewer | xargs | tr ' ' ','`
+              echo "Requesting review from reviewers in ci.yaml file: $reviewer"
+            done
```
mporrato, line 77: The purpose of the `xargs | tr ' ' ','` part in my comment was to avoid the for loop and add all the reviewers in a sigle call to `gh`.
It's not very clear from the `gh` documentation but if you look at the examples, `--add-reviewer` accepts multiple usernames separated by commas, so you should be able to just do:

```shell
gh pr edit $(params.pr_url) --add-reviewer $(echo $CI_REVIEWERS | xargs | tr ' ' ','`)
```

or, if you prefer:

```
gh pr edit $(params.pr_url) --add-reviewer $(echo $CI_REVIEWERS | paste -d, -s)
```

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
```diff
@@ -0,0 +1,79 @@
...
+        - name: GITHUB_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: $(params.github_token_secret_name)
+              key: $(params.github_token_secret_key)
+      image: "$(params.pipeline_image)"
+      script: |
+        #! /usr/bin/env bash
+        set -e
+        # do not use -x to not leak github secrets into logs!
```
mporrato, line 59: We are not handling the github token directly anymore, you can re-enable the `-x` if you want.

mporrato (changes requested):

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
```diff
@@ -0,0 +1,79 @@
...
+
+        AUTHOR_IS_REVIEWER=$(cat $(results.author_is_reviewer.path))
+        CI_REVIEWERS=$(cat $(results.list_reviewers.path))
+        # requesting review from repo maintainers if there are no reviewers in ci.yaml or no ci.yaml file
+        if [ -z "$CI_REVIEWERS" ]; then
+          gh pr edit $(params.pr_url) --add-reviewer redhat-openshift-ecosystem/pipeline-maintainers
```
mporrato, line 65: redhat-openshift-ecosystem/pipeline-maintainers is a rather big group.
We should probably add a field to the `config.yaml` file in the root of the repo containing the list of maintainers and get the list from there.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
tomasbakk: I tried the suggestion (with some additional changes trying to make it work) but it does not work because when the reviewers key is not there, the indexing in yq would fail - was searching for that issue yesterday/few days ago and it has something to do with how content from yaml is then parsed to json during storing it in results and if key is not there, it just fails on the indexing, if i remember correctly. 
I had the same issue as in the previous iteration, that is the reason why i switched to greping the key directly and then going through the additional else condition. This way i made it work and i did not have any errors  - whether reviewers are missing or the entire ci.yaml file is missing.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-reviewer.yml
tomasbakk: i created the PR for the config change https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/27

tomasbakk:
@ezopezo @mporrato can you have a look now? thank you!

mporrato (approved):
Looks good!

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
```diff
@@ -120,12 +120,37 @@ spec:
...
+      taskRef:
+        name: get-ci-reviewer
+        kind: Task
+      runAfter:
+        - detect-changes
+      params:
+        - name: pipeline_image
+          value: "$(params.pipeline_image)"
+        - name: git_username
+          value: $(params.git_username)
```
ezopezo, line 134:
```suggestion
          value: "$(params.git_username)"
```

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
```diff
@@ -120,12 +120,37 @@ spec:
...
+        - name: pr_url
+          value: $(params.git_pr_url)
```
ezopezo, line 136:
```suggestion
          value: "$(params.git_pr_url)"
```

ezopezo (approved):

PR #484, opened 2023-08-28
Add mode option to IIB builds

Allda:
The mode option allows users to configure how the update graph will be updated. The option is set in ci.yaml configuration file and is passed to the IIB build. The default value for the mode is "replaces".

JIRA: ISV-3837


ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
```diff
@@ -251,6 +267,12 @@ spec:
         # Adding a commit hash as a suffix to the image tag
         - name: index-image-destination-tag-suffix
           value: "-$(params.git_commit)"
+        - name: upgrade-graph-mode
+          value: "$(tasks.read-config.results.upgrade-graph-mode)"
+        # Production IIB instance doesn't support the upgrade path feature yet.
+        # TODO: switch it to PROD when the feature is available
+        - name: iib_url
```
jedinym, line 274: Should we create a Task ticket for this?

operator-pipeline-images/operatorcert/entrypoints/index.py
```diff
@@ -113,15 +119,17 @@ def add_bundle_to_index(
     iib_url: str,
     indices: List[str],
     image_output: str,
+    mode: Optional[str] = None,
```
jedinym, line 122: Do we need a default parameter here if we always provide the option?

operator-pipeline-images/operatorcert/entrypoints/index.py
```diff
@@ -53,6 +53,12 @@ def setup_argparser() -> argparse.ArgumentParser:  # pragma: no cover
         default="",
         help="A tag suffix to append to the index image when copying to the destination.",
     )
+    parser.add_argument(
+        "--mode",
+        choices=["replaces", "semver", "semver-skippatch"],
+        default="",
```
jedinym, line 59: I think having the default as `None` would be a bit cleaner, but your call.

jedinym (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
Allda: I am in sync with a iib developer who takes care of that. I'll change it once it is ready.

operator-pipeline-images/operatorcert/entrypoints/index.py
Allda: Fixed

operator-pipeline-images/operatorcert/entrypoints/index.py
Allda: Fixed

PR #486, opened 2023-08-30
[ISV-3824] Merge the PR if all tests are successful.

haripate:


ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
```diff
@@ -275,6 +275,28 @@ spec:
         - name: github_token_secret_key
           value: "$(params.github_token_secret_key)"
 
+    # merge PR
+    - name: merge-pr
+      runAfter:
+        - wait-prow-tests
+      taskRef:
+        name: merge-pr
```
Allda, line 283: The task checks if an auto-merge flag in the operator config is set before merging. @mporrato mentioned this is not needed for the community. Can you please sync and agree on whether this is fine or should be modified? Also, the ACs in the ticket say the task needs to verify if the PR is approved but this task approves it before merging.

mporrato: Indeed, community operators do not have the the `.merge` field in ci.yaml and many operators do not even have a ci.yaml file, so the merge-pr task as it is now would always fail.
For the approval from the bot, it's not how it works now but it won't affect the workflow so it's not a real issue if we keep that behaviour.
You could slightly alter the existing task to accept an additional `force_merge` parameter and correctly deal with a missing ci.yaml or a missing `.merge` filed.
You also need to make the execution of this task dependent on the result from ISV-3494 (using a tekton `when` clause).

Allda (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
```diff
@@ -327,6 +327,34 @@ spec:
         - name: github_token_secret_key
           value: "$(params.github_token_secret_key)"
 
+    # merge PR
+    - name: merge-pr
+      when:
+        - input: "$(tasks.get-ci-reviewer.results.author_is_reviewer)"
+          operator: notin
+          values: ["false"]
```
mporrato, line 335: Just for the sake of the principles of defensive programming I would use the opposite:

```suggestion
        - input: "$(tasks.get-ci-reviewer.results.author_is_reviewer)"
          operator: in
          values: ["true"]
```

but it's not a big deal here.

mporrato (approved):
Just a minor nitpick, but overall looks good!

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
haripate: Updated.

PR #494, opened 2023-09-06
Make --mode optional argument for iib build

Allda:
The mode can be only used by community pipeline and excluded for other pipelines. This change makes the option optional and only used when defined by the pipeline argument.


mantomas (approved):
Makes sense, LGTM 👍🏻

PR #498, opened 2023-09-12
[ISV-3982] create entrypoint for prow community

tomasbakk:
PR to community preprod main.sh: https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/72


tomasbakk:
@ezopezo can you please review this and let me know if this is the correct approach? thanks

operator-pipeline-images/operatorcert/community_prow/start_community_prow.py
```diff
@@ -0,0 +1,11 @@
+"""
+Entrypoint module for community-pipelines prow tests
+"""
+
+
+def initiate_community_prow(*args, **kwargs):
```
ezopezo, line 6: I would rather see this named as `main` (also file) invoked with `if __name__ == "__main__": main()`. Function does not have t o return anything. Also I would like to see here some log message proving that this is really triggered from community-operators-pipeline-preprod repo. After this, this have to be merged (also change on community side and tested, if log message is correctly displayed in prow logs. I appreciate also  @mporrato opinion about delegating this into separate repository grouping here all prow-side code logic.

tomasbakk: changed the naming and included logging

mporrato:
Just for the sake of demonstrating that we received the necessary information from prow and that we can talk to the openshift api from this script, can you add a function that logs the contents of the `PULL_REFS` environment variable and uses the [openshift-python-client](https://github.com/openshift/openshift-client-python) library to call something simple like `oc.get_server_version()` and logs the result? It will soon go away but it will be handy during initial testing.

tomasbakk:
> Just for the sake of demonstrating that we received the necessary information from prow and that we can talk to the openshift api from this script, can you add a function that logs the contents of the `PULL_REFS` environment variable and uses the [openshift-client-python](https://github.com/openshift/openshift-client-python) library to call something simple like `oc.get_server_version()` and logs the result? It will soon go away but it will be handy during initial testing.

tested the logs against my container image and both pull refs and server version were logged

Screenshot from 2023-09-21 21-53-27
Screenshot from 2023-09-21 21-52-16

tomasbakk:
@mporrato logs from pull_refs and server version are propagated in test_results (screens attached above). can you please re-review this? thanks!

mporrato (approved):

PR #500, opened 2023-09-13
Fix global variable usage in signing module

Allda:
After mypy and pylint fixes there is a bug which uses 2 variables with a same name. This commit fixes the bug.


mantomas (approved):
LGTM 👍🏻

PR #501, opened 2023-09-13
[ISV-3980] detect-changed-operators handles incorrectly rebased PRs

mporrato:
The detect-changed-operators entry point now uses the github api to get the list of affected files instead of checking out the head of the base branch from the local clone.

### DIFF
--- a/operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
+++ b/operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
@@ -5,7 +5,8 @@ import logging
 import os
 import pathlib
 import re
-from typing import Tuple, Optional, Set, Dict, List
+from dataclasses import dataclass, field
+from typing import Optional
 from operator_repo import Repo as OperatorRepo
 
 from github import Auth, Github
@@ -85,7 +86,7 @@ def github_pr_affected_files(pr_url: str) -> set[str]:
     return {x.filename for x in gh_pr.get_files()}
 
 
-def _find_owner(path: str) -> Tuple[Optional[str], Optional[str]]:
+def _find_owner(path: str) -> tuple[Optional[str], Optional[str]]:
     """
     Given a relative file name within an operator repo, return a pair
     (operator_name, bundle_version) indicating what bundle the file belongs to.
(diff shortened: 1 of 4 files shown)


operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
```diff
@@ -25,36 +26,104 @@ def setup_argparser() -> argparse.ArgumentParser:
...
     parser.add_argument("--verbose", action="store_true", help="Verbose output")
 
     return parser
 
 
-def detect_changed_operators(  # pylint: disable=too-many-locals,too-many-branches,too-many-statements
-    repo_path: str,
-    head_commit: str,
-    base_commit: str,
+def github_pr_affected_files(pr_url: str) -> set[str]:
```
Allda, line 55: There is already a function in `__init__.py` `get_files_added_in_pr`. Please consider reusing it or at least replacing it with your solution.

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
```diff
@@ -78,123 +147,76 @@ def detect_changed_operators(  # pylint: disable=too-many-locals,too-many-branch
...
+    modified_operators: Set[str] = set()
+    deleted_operators: Set[str] = set()
+    added_bundles: Set[Tuple[str, str]] = set()
+    modified_bundles: Set[Tuple[str, str]] = set()
+    deleted_bundles: Set[Tuple[str, str]] = set()
+
+    head_repo = OperatorRepo(repo_path)
+    base_repo = OperatorRepo(base_path)
+
+    for (  # pylint: disable=too-many-nested-blocks
```
Allda, line 189: Do you think this can be refactored or split into multiple functions that doesn't violate pylint? With a function that long, with so many variables, and nested conditions it is quite hard to read the code.

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
mporrato: I'm aware of that function, but it contains more logic than a simple "get me a list of affected files" and it's used in 10 different spots across the code base. I think that such a refactoring is way out of scope for this PR.

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
mporrato: I don't think we can get rid of the nested IFs because that's the core logic of the whole module, but I can try to break the main function in two or three parts to simplify it a bit.

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
Allda: ok, make sense

Allda (approved):

ezopezo (approved):

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
ezopezo: That looks really better :+1:

PR #507, opened 2023-09-25
updating the acquire-lease logic to add retries and timeouts

acornett21:
### Motivation
After seeing multiple issues where the release pipeline, is affectively single threaded and only can support 1 pull request (release) per hour, I started looking to see what configurations were available. 

### Details
The below changes are based on 'napkin math' that was observed when manually having to re-run release pipelines for an entire day. These are just meant to be a conversation starter about what can/should be improved, in the release pipeline. But ultimately, I would like at least `retries` to trickle down to all tasks that could have a timeout, that yields a manual re-run.

### Why Here?
Ideally, I'd love for downstream of the release pipeline to offer higher throughput, but that doesn't seem like it will happen, so hopefully we can make some improvements here that will lead to a better quality of life for ourselves and our users.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
@@ -83,3 +83,4 @@ spec:
         create_lease_yaml
         # Acquire Lease
         acquire_lease
+      timeout: "0h35m0s"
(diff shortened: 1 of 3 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/acquire-lock.yaml
```diff
@@ -83,3 +83,4 @@ spec:
         create_lease_yaml
         # Acquire Lease
         acquire_lease
+      timeout: "0h35m0s"
```
Allda, line 86: This timeout might be in conflict with a timeout given by the task parameter. Is this needed if the task itself waits for a lease with a timeout already?

Allda (approved):

PR #509, opened 2023-09-27
[ISV-4091] Align release pipeline to the new version of detect-changes

mporrato:
This fixes the usage of the detect-changes task in the release pipeline that I forgot to update when fixing the task behaviour.
Also update the input arguments of the git-clone tasks to be more specific and use the commit sha instead of the pr branch ref.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/community-release-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/community-release-pipeline.yml
@@ -104,7 +104,7 @@ spec:
         - name: env
           value: $(params.env)
 
-    # clone repository
+    # Clone the repository contents from the PR branch
     - name: clone-repository
       taskRef:
         name: git-clone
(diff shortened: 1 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/pipelines/community-release-pipeline.yml
```diff
@@ -113,11 +113,9 @@ spec:
         - set-env-community
       params:
         - name: url
-          value: $(params.git_repo_url)
+          value: "$(params.git_repo_url)"
```
ezopezo, line 116: Is this valid that this is utilizing a different parameter than in hosted pipeline?

mporrato: We don't have the PR branch URL as an input to the release pipeline; we could add it but, since this is called *after* the PR has been merged, the sha for the head of the PR will be present in the destination branch too.
I tested it and it seems to work fine.

ezopezo: Makes sense, thanks for the clarification!

ezopezo (approved):

tomasbakk (approved):

PR #513, opened 2023-10-02
ISV-4029: remove obselete/unused CA

johnbe11:


johnbe11:
@mporrato PTAL

johnbe11:
@mporrato fyi - rebased to pick up your CVE fix

johnbe11:
@mporrato one is already expired, and the other is about to expire.  Neither of them are used/referenced by the project.

mporrato:
> @mporrato one is already expired, and the other is about to expire. Neither of them are used/referenced by the project.

It's being copied into the container image in https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/d166012e2b7a680d2802e56b57b3b6b5428c4b51/operator-pipeline-images/Dockerfile#L14 but I see we already have the 2022 CA cert so we should be fine :crossed_fingers:

PR #518, opened 2023-10-09
WIP [ISV-3979] Properly handle /test commands

jedinym:


Allda:
@jedinym As we already discussed on the status update this PR won't be needed if we drop prow jobs completely. I created a separate ticket for the re-trigger mechanism. https://issues.redhat.com/browse/ISV-4145

PR #520, opened 2023-10-12
[ISV-4082] Changes not affecting any bundle should not trigger tests

mporrato:


ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
```diff
@@ -409,6 +428,30 @@ spec:
         - name: github_token_secret_key
           value: "$(params.github_token_secret_key)"
 
+    - name: skip-prow-tests
```
Allda, line 431: This can stay here for now, but it will be most likely removed when adding a preflight-trigger to the pipeline.

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-release-pipeline.yml
```diff
@@ -173,17 +174,47 @@ spec:
           workspace: repository
           subPath: base
 
+    # acquire/lease the resource to resolve the conflict of concurrent pipelineruns
+    - name: acquire-lease
```
Allda, line 178: Is it really necessary to acquire a lease at the beginning of the pipeline? Is there any other critical section other than IIB build?

mporrato: I moved it here just to ensure that the release of the lease in the `finally` section has a matching acquire task even when we skip the actual release steps, but I'm not sure it's required. And the tasks between this and the start of the actual critical section are just a couple and quite simple.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-paths-from-name-and-version.yml
```diff
@@ -0,0 +1,39 @@
+---
+# Given an operator name and bundle version, return the path to the operator and bundle directories
+apiVersion: tekton.dev/v1beta1
```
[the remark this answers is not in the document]
mporrato: Good idea. I'll do that.

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-release-pipeline.yml
Allda: The delete lease is implemented like `oc delete lease $LEASE_NAME --ignore-not-found=true` so I think it should not be a problem. My concern about moving the acquire task to the early stages of the pipeline is that we already faced an issue in ISV release pipeline when the lease caused a bottleneck when multiple PRs have been merged simultaneously. We should aim to have the critical section as small as possible.

mporrato: My concern is that when we get to the `oc delete` step without acquiring the lease first, it could delete the lease acquired by another pipeline run: I don't see any mechanism in place to avoid that.
Anyway, I've moved the task back to its original position and added a `when` clause to the lease release task but still need to iron out a remaining issue. Will ping you when it's ready for another look.

Allda (approved):

PR #521, opened 2023-10-13
Re-trigger pipelines using labels

Allda:
The label re-trigger mechanism is reused for community pipelines and each pipeline can be triggered by corresponding label on PR.

JIRA: ISV-4145


haripate (approved):
LGTM

PR #527, opened 2023-10-23
Add static check for bundle graph check

Allda:
Two new test are introduced for a community pipeline:
 - Detection of dangling bundles in a channel
 - Detection of loops in channel

JIRA: ISV-3795

### DIFF
--- a/operator-pipeline-images/operatorcert/static_tests/community/bundle.py
+++ b/operator-pipeline-images/operatorcert/static_tests/community/bundle.py
@@ -10,6 +10,7 @@ import json
 import re
 import subprocess
 from collections.abc import Iterator
+from typing import Any, List
 
 from operator_repo import Bundle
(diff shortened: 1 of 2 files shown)


mporrato (approved):
Looks good to me. The only nitpick is that, since this test validates the dependencies between bundles of a given operator, I'd say it conceptually belongs more to the operator checks module than the bundle checks module.

Allda:
> Looks good to me. The only nitpick is that, since this test validates the dependencies between bundles of a given operator, I'd say it conceptually belongs more to the operator checks module than the bundle checks module.

Initially, I placed it into the operator module, but then I realized the new pipeline only checks the current operator that is submitted so I moved it to the bundle module and always start from the submitted bundle.

piyush-nimbalkar:
Ah, that was indeed the issue. Due to some issue in our automation we did the release manually this time and messed up.

It works now after rebasing. Thank you for looking at it!

PR #528, opened 2023-10-24
Fix cert-project-check task

Allda:
When the project is missing in the ci.yaml file the error message was not raised and process continued with a project ID set to "null".

This commit fixes the check to detect the empty project id in config.


jedinym (approved):
lgtm

PR #529, opened 2023-10-24
[ISV-4109] Introduction of Hadolint and Ansible-lint.

BorekZnovustvoritel:
Preparation for the introduction of Pymarkdownlnt.


BorekZnovustvoritel:
Integration tests successfully ran, ready for review.

pyproject.toml
```diff
@@ -84,6 +84,7 @@ operatorcert-dev = [
     "types-requests>=2.31.0.2",
     "pylint>=2.17.5",
     "openshift-client>=1.0.19",
+    # "pymarkdownlnt>=0.9.13.4",
```
Allda, line 87: Is this intentional change?

tox.ini
```diff
@@ -73,3 +75,20 @@ commands = pdm export \
 [testenv:pdm-lock-check]
 allowlist_externals = pdm
 commands = pdm lock --check
+
+[testenv:hadolint]
+allowlist_externals = hadolint
+groups = dev
+commands = hadolint -V --failure-threshold warning \
+           --info DL3006 --info DL3007 --info DL3013 --info DL3041 \
```
Allda, line 83: Can you please describe what are those `DL` rules and why we need to specifically make them as info?

BorekZnovustvoritel: These DL all relate to pinning versions in different package managers. See [https://github.com/hadolint/hadolint/wiki/DL3006](DL3006: include image tag), [https://github.com/hadolint/hadolint/wiki/DL3007](DL3007: warning against using `:latest` tag), [DL3013: Pip package version pinning] (https://github.com/hadolint/hadolint/wiki/DL3013) and [DL3041: dnf version pinning](https://github.com/hadolint/hadolint/wiki/DL3041). I wanted to keep the messages there as a *best practice*, but did not want to fail the validation because I feel like using latest images and packages is the entire purpose of building images on every push.

Shall I:
- Actually pin the versions
- Write comments describing why this step was done or
- Leave as-as?

pyproject.toml
BorekZnovustvoritel: I left that there intentionally, but it can be removed if you prefer cleaner pyproject.toml file. Shall I do that?

tox.ini
Allda: When I tried to run on my computer I only see `DL3041` and `DL3013` as a warning. I am fine with keeping those as indo, but are you sure other are also needed?

BorekZnovustvoritel: You are correct, they aren't actually needed. I will remove them.

Allda (approved):

ansible/roles/config_ocp_cluster/tasks/main.yml
```diff
@@ -1,10 +1,11 @@
 ---
 - name: Customize TektonConfig
-  k8s:
+  kubernetes.core.k8s:
```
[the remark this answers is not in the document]
BorekZnovustvoritel: I changed it now and also I have changed the version to the [latest release of the package (2.4.0)](https://galaxy.ansible.com/ui/repo/published/kubernetes/core/). If there is a specific reason not to do that, let me know. Thank you!

acornett21:
@Allda @BorekZnovustvoritel Are the folder name changes necessary for the linter, can't they be ignored? There are downstream consumers (and documentation) and changing the folder names make implementing failover/fallback logic to support both folder structures less the ideal.

BorekZnovustvoritel:
The issue I have encountered with the linter is that if the role dir does contain the `-` character, then it does not apply all the role-related rules to the directory even though it is located in the `ansible/roles` directory.

From my testing, if we were to rollback the changes, the linter would mark the code as passing and would be less enforcing. I don't think there is a way to keep the role name and have the linter recognize the directory as a role without failing. The related linter rule can be found [here](https://ansible.readthedocs.io/projects/lint/rules/role-name/).

If the changes are breaking and have to be rolled back, let me know. I did not expect this change to propagate outside of this repository.

PR #531, opened 2023-10-24
fix: remove additional parameter `--bundle-path` since is not needed

### LINKED ISSUE #530: ocp-version-info command fails with a wrong argument
After a couple of tests we found out that the task `operator-ci-pipeline-run-get-supported-versions-pod` fails with the following error:

```
kubectl -n oco logs operator-ci-pipeline-run-get-supported-versions-pod -c step-supported-version-check
++ cat config.yaml
++ yq -r .organization
+ ORGANIZATION=community-operators
+ OCP_VERSIONS_OUTPUT_FILE=./ocp_versions.json
+ '[' -n operators/cloudnative-pg/1.21.0 ']'
+ ocp-version-info --output-file ./ocp_versions.json --bundle-path operators/cloudnative-pg/1.21.0 community-operators
usage: ocp-version-info [-h] [--pyxis-url PYXIS_URL]
                        [--output-file OUTPUT_FILE] [--verbose]
                        bundle_path
                        {certified-operators,redhat-marketplace,community-operators}
ocp-version-info: error: unrecognized arguments: --bundle-path
```
Looks like this will be due to this commit: https://github.com/redhat-openshift-ecosystem/operator-pipelines/commit/a53683b37538f9bfdcbd18eb3867773856ceb2a0 
And this change:
```
-        ocp-version-info \
-          --output-file $OCP_VERSIONS_OUTPUT_FILE \
-          $(params.bundle_path) \
-          $ORGANIZATION
+        if [ -n "$(params.bundle_path)" ] ; then
+            ocp-version-info \
+              --output-file $OCP_VERSIONS_OUTPUT_FILE \
+              --bundle-path "$(params.bundle_path)" \
+              $ORGANIZATION
+        else
+            ocp-version-info \
+              --output-file $OCP_VERSIONS_OUTPUT_FILE \
+              $ORGANIZATION
+        fi
```

There's any chance this is an actual bug or we should change something in the arguments when creating these steps?

Regards

sxd:
Closes #530

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/get-supported-versions.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/get-supported-versions.yml
@@ -47,7 +47,7 @@ spec:
         if [ -n "$(params.bundle_path)" ] ; then
             ocp-version-info \
               --output-file $OCP_VERSIONS_OUTPUT_FILE \
-              --bundle-path "$(params.bundle_path)" \
+              "$(params.bundle_path)" \
               $ORGANIZATION
         else
             ocp-version-info \


sxd:
Hi @mporrato 

can you take a look into this PR?

Regards!

Allda:
The `--bundle-path` is a valid argument for the scrip as you can see here: https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/operator-pipeline-images/operatorcert/entrypoints/ocp_version_info.py#L24

The fact that you are getting the error in https://github.com/redhat-openshift-ecosystem/operator-pipelines/issues/530 is mostly likely due to a mismatch between the release of the Tekton pipeline and python scripts. I suggest you always use the latest official release instead of pulling pipelines from the `main` branch.

The new version with this flag should be released later today.

sxd:
Hi @Allda 

The problem is that using the https://github.com/redhat-openshift-ecosystem/operator-certification-operator we cannot select a tag but only branches, so I think I'll ask to the operator-certification-operator team if we can add this as a new feature so we can use only one specific version of operator-pipelines and not pull from only main

Regards!

PR #543, opened 2023-11-06
Fix role name after refactoring

Allda:
The role name uses underscore instead of the dash. This caused the role execution was skipped from the playbook.


mantomas (approved):

BorekZnovustvoritel (approved):
LGTM

PR #550, opened 2023-11-13
Add new task to request signature of blobs

ernesgonzalez33:
- Add request-signature-blob task to be able of using request-signature to sign a blob
- Modify request-signature.py to sign container images and blobs depending of the arguments passed


Allda (approved):
The changes look fine to me. There are a couple of linting errors. Please review the output of the CI and fix the violations. I'll merge it when the pipeline passes.

Allda:
> > The changes look fine to me. There are a couple of linting errors. Please review the output of the CI and fix the violations. I'll merge it when the pipeline passes.
> 
> @Allda this should be fixed! Could you please approve the checks to run?

There are still 2 CI errors, one is the black format (you can fix it by running `tox -e black-format`). And the second failure comes from unit test coverage.

ernesgonzalez33:
@Allda should be solved now. I had to reorder things a bit in the function because the tests were going through a place they shouldn't have gone in the first place

PR #552, opened 2023-11-16
[ISV-3786] create api deprecation test

tomasbakk:


operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -34,10 +34,53 @@ class GraphLoopException(Exception):
     """
 
 
+def extract_ocp_version_from_bundle_metadata(ocp_metadata_version: Any) -> Any:
```
Allda, line 37: The `com.redhat.openshift.versions` can be defined in more formats than just a simple version or range. Please make sure all possible formats are supported or get the supported version form Pyxis based on the annotation.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -34,10 +34,53 @@ class GraphLoopException(Exception):
...
+        "v4.7": "1.20",
+        "v4.8": "1.21",
+        "v4.9": "1.22",
+        "v4.10": "1.23",
+        "v4.11": "1.24",
+        "v4.12": "1.25",
+        "v4.13": "1.26",
+        "v4.14": "1.27",
+        "v4.15": "1.28",
+    }
```
mporrato, line 72: This piece of information will be needed by other tests (at least ISV-4290). Can you move it to a place where it would be easier to share between multiple tests?

mporrato (changes requested):

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
tomasbakk: I grepped the community and certified repos for metadata annotations and found out that equal sign is also used to annotate single ocp version, so I included additional check here.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
tomasbakk: table was moved from the test

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
Allda: @tomasbakk The version definition in this field is more than just a static selection of a single version. As a user, you can define multiple ways and each version format has a different meaning. That's why I suggest using a Pyxis to determine the latest supported version based on the annotation.

For example when a version annotation defines `v4.10` it means it supports all versions from `4.10` up to the latest supported one which is currently `4.14`.

Here is the query that gives you a supported version based on the annotation. From that list, you will use the last version available.
https://catalog.redhat.com/api/containers/v1/operators/indices?filter=organization==certified-operators&ocp_versions_range=v4.10

If you wan to read more about this annotation here is a detailed documentation: https://docs.engineering.redhat.com/display/CFC/Delivery

tomasbakk: aha, thank you! i thought that users have more options how to annotate the version, but in the end what matters is the upper ocp range (if range is mentioned) or single definition. now that you mentioned it, i remember that we discussed the supported versions during some meeting past week, and it makes sense to me now. i will go through the documentation and fix it. thanks.

tomasbakk: @Allda I added the call to get the indices and parse the OCP version based on that. Can you please check it out if it is alright? Thanks!

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -27,17 +28,66 @@
...
+    # to run API deprecation test
+    indices_url = "https://catalog.redhat.com/api/containers/v1/operators/indices"
+    params = {
+        "filter": "organization==community-operators",
+        "sort_by": "ocp_version[desc]",
+        "include": "data.ocp_version",
+        "ocp_versions_range": ocp_metadata_version,
+    }
+    # timeout param set in order to not hang indefinitely
+    rsp = requests.get(indices_url, params=params, timeout=30)
```
Allda, line 73: We should handle API error codes and fail with a clear message in case the API is not responding correctly.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -27,17 +28,66 @@
...
+    """
+    # we are calling indices to extract the supported OCP version
+    # which is then used in operator-sdk bundle validate
+    # to run API deprecation test
+    indices_url = "https://catalog.redhat.com/api/containers/v1/operators/indices"
+    params = {
+        "filter": "organization==community-operators",
+        "sort_by": "ocp_version[desc]",
+        "include": "data.ocp_version",
+        "ocp_versions_range": ocp_metadata_version,
```
Allda, line 70: Given you care only about the latest version you can limit the page size to 1 item here.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -27,17 +28,66 @@
...
+    ocp_version = get_indices[0].get("ocp_version") if get_indices else "4.16"
+
+    return ocp_version
+
+
 def run_operator_sdk_bundle_validate(
     bundle: Bundle, test_suite_selector: str
 ) -> Iterator[CheckResult]:
     """Run `operator-sdk bundle validate` using given test suite settings"""
+    ocp_annotation = bundle.annotations.get("com.redhat.openshift.versions", {})
```
Allda, line 85: The default value here should not be an empty dict. Since we expect a string from this annotation I suppose the better value would be an empty string or none.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
Allda: Also in case the annotation is missing you should exclude the `ocp_versions_range` which gives you the latest supported version.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -27,17 +28,66 @@
...
+        "filter": "organization==community-operators",
+        "sort_by": "ocp_version[desc]",
+        "include": "data.ocp_version",
+        "ocp_versions_range": ocp_metadata_version,
+    }
+    # timeout param set in order to not hang indefinitely
+    rsp = requests.get(indices_url, params=params, timeout=30)
+    get_indices = rsp.json().get("data")
+
+    ocp_version = get_indices[0].get("ocp_version") if get_indices else "4.16"
```
Allda, line 76: The default here is I think not correct. If the API doesn't return anything the test should fail in my opinion.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -47,6 +97,7 @@ def run_operator_sdk_bundle_validate(
         bundle.root,
         "--select-optional",
         test_suite_selector,
+        f"--optional-values=k8s-version={kube_version_for_deprecation_test}",
```
Allda, line 100: How does the new parameter communicate the deprecated API? Is this part of the structured json output or is it just a warning log?

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
tomasbakk: I was thinking about it, based on the info in jira, if target version is not specified, test should be executed against all supported versions. Since that would be IMO unnecessary I wanted to execute test against the latest OCP version we have. What we can do if the user does not specify annotation is either fail the test entirely or fallback to the latest supported version and test it against it. I think you are right and strict approach might be better.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
tomasbakk: it is part of a json output

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
tomasbakk: Based on the discussion via chat and also as stated in [ISV-4290](https://issues.redhat.com/browse/ISV-4290) in case the annotation is missing, we would test all supported versions - which means 4.16:1.29 (current OCP:k8s) version would be used as a default value for the test. The test is executed retrospectively, so it would test also all past deprecations/supported ocp versions.

operator-pipeline-images/tests/static_tests/community/test_bundle.py
```diff
@@ -13,10 +13,33 @@
...
 from tests.utils import bundle_files, create_files, merge
 
+from requests import HTTPError
 
+
+@pytest.mark.parametrize(
+    ["ocp_version", "expected"],
+    [("4.8-4.12", "4.12"), ("=v4.9", "4.9"), (None, "4.16")],
+)
+def test_process_ocp_version(ocp_version: str, expected: str) -> None:
```
Allda, line 27: A test should not call real API. Please mock Pyxis API here.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -27,17 +29,76 @@
...
+    rsp = requests.get(indices_url, params=params, timeout=60)  # type: ignore[arg-type]
+
+    try:
+        rsp.raise_for_status()
+    except requests.HTTPError as exc:
+        LOGGER.error("GET request to fetch the indices failed with: %s", exc)
+        return None
+
+    get_indices = rsp.json().get("data")
+    ocp_version = get_indices[0].get("ocp_version")
```
Allda, line 87: There is no guarantee that Pyxis will return one item. The list might be empty in case a range is pointing to some old EOL version. Return None is fine here.

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -27,17 +29,76 @@
...
+
 def run_operator_sdk_bundle_validate(
     bundle: Bundle, test_suite_selector: str
 ) -> Iterator[CheckResult]:
     """Run `operator-sdk bundle validate` using given test suite settings"""
+    ocp_annotation = bundle.annotations.get("com.redhat.openshift.versions", None)
+
+    ocp_version_to_convert = process_ocp_version(ocp_annotation)
+
+    kube_version_for_deprecation_test = K8S_TO_OCP.get(ocp_version_to_convert)
```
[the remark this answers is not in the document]
tomasbakk: I tested this case along with wrong pyxis response; In case None was returned, operator-sdk executed the test against the None flag, here are the logs: https://gist.github.com/rh-operator-bundle-bot/65d6d2c0f79400ef2ddb8afb0267cc2f
But I can change it. Do you mean to omit k8s-version argument from the test entirely when the return from process_ocp_version() is None?

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
tomasbakk: Added None as fallback value. Also, if the annotation points to a specific EOL version (=v4.10 and down), pipeline will fail on get-supported-version task, due to EOL version, and will not reach the static-tests.

Allda (approved):

PR #554, opened 2023-11-23
Set github PR title with operator name and version

Allda:
The community hosted operator pipeline sets PR title based on an added operator. All PR titles will be formatted with operator name and version.

JIRA: ISV-4262

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-pr-title.yaml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-pr-title.yaml
@@ -1,0 +1,45 @@
+---
+apiVersion: tekton.dev/v1beta1
+kind: Task
+metadata:
+  name: set-github-pr-title
+spec:
+  params:
+    - name: pipeline_image
+      description: The common pipeline image.
+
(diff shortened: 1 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/set-github-pr-title.yaml
```diff
@@ -0,0 +1,45 @@
...
+        - name: GH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: $(params.github_token_secret_name)
+              key: $(params.github_token_secret_key)
+      script: |
+        #! /usr/bin/env bash
+        set -ex
+
+        if [[ $(params.bundle_version) == "" ]]; then
```
ezopezo, line 40: Task is not triggered in pipeline, when detect-changes does not identify added bundle, so is this condition even needed?

Allda: I want to keep it there in case somebody accidentally updates the pipeline definition which can cause unexpected issues.

ezopezo (approved):

PR #557, opened 2023-11-28
Allow skipping preflight tests using PR labels

Allda:
By adding a label to a PR in format tests/skip/{test_name} admins can skip certain tests and make it optional.

JIRA: ISV-4316

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/evaluate-preflight-result.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/evaluate-preflight-result.yml
@@ -62,11 +62,24 @@ spec:
         #! /usr/bin/env bash
         set -e -o pipefail
 
+        EXTRA_ARGS=""
+
+        # Get labels from the pull request
+        gh pr view "$(params.request_url)" --json labels > labels.json
+
+        # Following command filters out labels that start with prefix and merge them to single string separated by comma
+        SKIP_TESTS=`jq -r '.labels[] | select(.name | startswith("tests/skip")) | .name' labels.json | paste -sd "," |sed 's/tests\/skip\///g'`
+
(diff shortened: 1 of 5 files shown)


mantomas (approved):
LGTM

ansible/roles/operator-pipeline/templates/openshift/tasks/evaluate-preflight-result.yml
```diff
@@ -53,7 +53,12 @@ spec:
       workingDir: $(workspaces.input.path)
       image: "$(params.pipeline_image)"
       env:
-        - name: GITHUB_TOKEN
+        - name: GITHUB_TOKEN # Token used by internal CLI tool
+          valueFrom:
+            secretKeyRef:
+              name: $(params.github_token_secret_name)
+              key: $(params.github_token_secret_key)
+        - name: GH_TOKEN # Token used by the GH tool
```
mporrato, line 61: Is this really needed? According to the [docs](https://cli.github.com/manual/gh_help_environment), the `gh` tool should work with either `GITHUB_TOKEN` or `GH_TOKEN`.

mporrato (approved):
LGTM, but this won't have the same behaviour as the current pipeline because it will require manual interaction.
We should probably add a new task at the beginning of the pipeline to automatically apply some skip-labels based on the pr contents (for example extract the allow-list from the existing `ci/pipeline-config.yaml` or perhaps adding something similar to `config.yaml` and check if the affected operator matches any of the names), but this can be done in another PR.

ansible/roles/operator-pipeline/templates/openshift/tasks/evaluate-preflight-result.yml
Allda: Ah, good point. I removed the this section.

PR #569, opened 2024-01-08
Add task that automatically skips tests for exceptions

Allda:
A configuration file determines which tests can be skipped. The task introduced in this commit reads the config and applies labels if test exception matches with submitted operator.

JIRA: ISV-4382

### DIFF
--- a/operator-pipeline-images/operatorcert/entrypoints/apply_test_waivers.py
+++ b/operator-pipeline-images/operatorcert/entrypoints/apply_test_waivers.py
@@ -1,0 +1,124 @@
+"""Configure test suite for given operator by adding/removing labels"""
+import argparse
+import logging
+import os
+import re
+from typing import Any, Dict
+
+from github import Auth, Github
+from operatorcert.github import add_or_remove_labels
+from operatorcert.logger import setup_logger
+from operatorcert.utils import get_repo_config
+
+LOGGER = logging.getLogger("operator-cert")
+
+
+def setup_argparser() -> argparse.ArgumentParser:
+    """
+    Setup argument parser
+
+    Returns:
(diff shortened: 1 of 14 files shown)


mporrato (approved):
Not thrilled by the choice of "test-config" as the task and script name. Perhaps "apply-test-waivers" or something like that would be more descriptive.
Other than that, the logic looks solid to me. :+1:

Allda:
> apply-test-waivers

That makes sense. I changed the name based on your suggestion. Thanks!

operator-pipeline-images/operatorcert/entrypoints/apply_test_waivers.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Returns:
+        bool: True if test can be ignored, False otherwise.
+    """
+    test_name = test.get("name")
+    LOGGER.debug("Checking test %s", test)
+
+    for ignored_operator in test.get("ignore_operators", []):
+        LOGGER.debug("Checking ignored operator %s", ignored_operator)
+        if not re.match(ignored_operator, operator_name):
```
jedinym, line 72: It might not matter in this case, but do we want to match only a substring of the test name here?

Allda: Our team will define these regex exceptions so we have control over what's matched. Is there any concern that this implementation might not cover?

Allda: You can see real config with regexes here: https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/3846/files

jedinym (approved):
LGTM

PR #572, opened 2024-01-12
Turn annotations validation to static check

Allda:
The annotations-validation tekton task migrated to a static check. The static check brings several benefits:
 - testability
 - reporting results to PR
 - simplification of a pipeline

JIRA: ISV-4392


mantomas (approved):
LGTM

PR #575, opened 2024-01-17
[ISV-4374] Update the validation for metadata.annotations.containerImage field.

haripate:


operator-pipeline-images/operatorcert/static_tests/community/bundle.py
```diff
@@ -175,7 +175,9 @@ def check_required_fields(bundle: Bundle) -> Iterator[CheckResult]:
         ("metadata.annotations.categories", validate_categories, False),
         (
             "metadata.annotations.containerImage",
-            re.compile(r"[^/]+/[^/]+/[^/:]+:.+"),
+            re.compile(
+                r"^([\w.-]+(?:/[\w.-]+)/)?([\w.-]+(?:[._-][\w.-]+)/)?([\w.-]+):([\w.-]+)$"
```
mporrato, line 179: You don't need the `^` prefix and `$` suffix here because we are using `re.match`, as opposed to `re.find`.

I'm afraid using `[\w.-]` might be too strict: for example it won't match something like `example.com/~foo/bar:baz`: the `~` is for sure a valid character in an URL so I would guess some container registry implementation might do the same. Also this won't work when the pullspec uses a sha256 instead of a simple tag (i.e.: `example.com/~foo/bar@sha256:e19a...`).

You don't need to mark groups as non-capturing (i.e. use `?:` after `(`) because we are just checking for a full match.

Have you considered just changing `r"[^/]+/[^/]+/[^/:]+:.+"` into `r"([^/]+/){2,}[^/:]+:.+"`?

mporrato (changes requested):
Can you please add a simple test case or update one of the existing ones to capture the fact that we need to support more than a single level of namespaces in container specs?

operator-pipeline-images/operatorcert/static_tests/community/bundle.py
haripate: Ah I see, I tried to implement the suggested regex, but I was grouping the wrong sequence here. Thank you, Maurizio, for the suggestion. I will update the regex now.

Allda: Can you please explain the regex to me? I think this is also too strict and for example doesn't support usecases like this: `foo.bar.com/foo:bar`

Allda: Also based on the regex format check the regex is not valid and is missing escaped slashes.

haripate: I already updated the regex as suggested by Maurizio. Can you please review that?

Allda: I was referring to the new latest regex in your code.

Allda: The OCI registry validates the repo name using the following regex. You can either use that one or at least don't require at least 2 `/` in regex.
https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests

Allda: And please add more test cases to cover multiple formats to prove the validation works fine.

Allda (approved):

PR #576, opened 2024-01-17
ISV-4298: override default timeout values for Import *-index imagestr…

johnbe11:
ISV-4298: override default timeout values for Import *-index imagestream tasks, to mitigate increased response times from import request

(timeout specified as integer seconds, default is 30)


johnbe11:
@Allda PTAL

Allda:
Can you please also add a tag to the ansible task and skip it during a deployment to real env? The import is only needed for integration tests.

Allda (changes requested):
As mentioned in the private chat this PR needs to be tested against real ocp instance to test if increased timeout is not ignored by the default timeout of OCP API.

Allda (approved):
Looks good to me :) Please confirm the timeout works as expected in parallel. I would like to merge this ASAP to no longer block releases.

PR #579, opened 2024-01-24
Publish resources using image requests

Allda:
The publishing process of metadata in Pyxis is handled by unified process that is common for both containers and operators.

The process required a bunch of refactoring and simplification in the pipeline. This commit removes all unused or duplicated steps and reduce pyxis interaction to only one task that handles everything.

JIRA: ISV-3989


operator-pipeline-images/operatorcert/entrypoints/create_container_image.py
```diff
@@ -129,16 +122,29 @@ def prepare_parsed_data(skopeo_result: Dict[str, Any]) -> Dict[str, Any]:
...
+
+    Args:
+        skopeo_result (Dict[str, Any]): Skopeo inspect result
+
+    Returns:
+        int: Image size in bytes
+    """
+    size = 0
+    for layer in skopeo_result.get("LayersData", []):
+        size += layer.get("Size", 0)
```
mporrato, line 137:
```suggestion
        size += layer.get("Size") or 0
```

operator-pipeline-images/operatorcert/pyxis.py
```diff
@@ -289,3 +290,82 @@ def get_repository_by_isv_pid(base_url: str, isv_pid: str) -> Any:
...
+
+    image_request = post(image_request_url, payload)
+
+    return image_request
+
+
+def wait_for_image_request(base_url: str, image_request_id: str) -> Any:
+    """
+    Wait for image request to finish and transition to one of the final states.
+    The request is tracked and status is printed every 5 seconds.
```
mporrato, line 331: Shouldn't we also have a deadline for this function to terminate? If, for some reason, the image request never transitions to one of the expected final statuses, we keep hammering pyxis...

ansible/roles/integration_tests/tasks/tools.yml
```diff
@@ -10,7 +10,7 @@
 
     - name: Download and extract the tkn binary
       ansible.builtin.unarchive:
-        src: https://mirror.openshift.com/pub/openshift-v4/clients/pipelines/1.10.0/tkn-linux-amd64.tar.gz
+        src: https://mirror.openshift.com/pub/openshift-v4/clients/pipelines/1.11.2/arm64/linux/tkn-linux-arm64.tar.gz
```
mporrato, line 13: I think you pasted the wrong URL here: this is the arm64 binary tarball. We need the amd64 version.

mporrato (changes requested):
Only one issue, with the arch of the tekton tarball and a few suggestions.

ansible/roles/operator-pipeline/templates/openshift/tasks/get-pyxis-certification-data.yml
```diff
@@ -89,14 +91,41 @@ spec:
 
         get_from_cp isv_pid '.container.isv_pid' | tee $(results.isv_pid.path)
         get_from_cp repository_name '.container.repository_name' | tee $(results.repo_name.path)
-        get_from_cp org_id '.org_id' | tee $(results.org_id.path)
         get_from_cp operator_distribution '.operator_distribution' | tee $(results.operator_distribution.path)
         get_from_cp project_status '.project_status' | tee $(results.project_status.path)
 
+        get_from_cp org_id '.org_id' > org_id.txt
+        cat org_id.txt | tee $(results.org_id.path)
```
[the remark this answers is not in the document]
Allda: Fixed

operator-pipeline-images/operatorcert/entrypoints/create_container_image.py
Allda: Fixed

operator-pipeline-images/operatorcert/pyxis.py
Allda: I added a timeout and configurable interval.

ansible/roles/integration_tests/tasks/tools.yml
Allda: Good catch! I fixed the URL

mporrato (approved):

PR #585, opened 2024-02-01
[ISV-4438] Improve static check of metadata.anotations.alm-examples

mporrato:
The current implementation of `check_required_fields()` performs a very simple check of the `metadata.anotations.alm-examples` field that does not prevent the submission of a bundle with an incorrect alm-examples (for example when it does not contain a valid json list) and, in cases where the operator does not provide any CRD, it returns a false positive.

Since this check is already performed by `operator-sdk bundle validate` checks, we can remove the field check from `check_required_fields()` and let operator-sdk do the validation.


Allda (approved):
I appreciate you looking at existing sdk instead of implementing the check by ourselves. :+1:  Could there be other tests already covered by the sdk?

mporrato:
> I appreciate you looking at existing sdk instead of implementing the check by ourselves. 👍 Could there be other tests already covered by the sdk?

Possibly. Would have to look at the operator-sdk code to find the overlaps. I will create a spike for that.

PR #587, opened 2024-02-07
Wait for freshness grades before publishing

Allda:
The publishing script waits for freshness grades to become available before submitting an image requires that publishes the image.

JIRA: ISV-4529

### DIFF
--- a/operator-pipeline-images/operatorcert/pyxis.py
+++ b/operator-pipeline-images/operatorcert/pyxis.py
@@ -386,3 +386,51 @@ def wait_for_image_request(
             )
             return image_request
         time.sleep(delay)
+
+
+def wait_for_container_grades(
+    base_url: str, image_identifier: str, timeout: float = 60 * 10, delay: float = 5
+) -> Any:
+    """
+    Wait for container freshness grades to become available for
+    the given image identifier.
+
+    Args:
+        base_url (str): Pyxis base URL
+        image_identifier (str): Container image identifier
+        timeout (float): Timeout in seconds
+        delay (float): Delay between status checks in seconds
+    """
+    container_grades_url = urljoin(base_url, f"/v1/images/id/{image_identifier}")
+    start_time = now()
+    while True:
+        resp = get(container_grades_url, params={"include": "freshness_grades,_id"})
+        try:
+            resp.raise_for_status()
+        except requests.HTTPError:
+            LOGGER.exception(
+                "Unable to get container image grades %s - %s - %s",
+                container_grades_url,
+                resp.status_code,
+                resp.text,
+            )
+            raise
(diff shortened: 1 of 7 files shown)


operator-pipeline-images/operatorcert/pyxis.py
```diff
@@ -386,3 +386,52 @@ def wait_for_image_request(
             )
             return image_request
         time.sleep(delay)
+
+
+def wait_for_container_grades(
+    base_url: str, image_identifier: str, timeout: float = 60 * 10, delay: float = 5
+) -> Any:
```
mantomas, line 393: Shouldn't this simply return None? I can see it returns `container_image`, is there reason for that except for testing purpose?

operator-pipeline-images/operatorcert/pyxis.py
```diff
@@ -386,3 +386,52 @@ def wait_for_image_request(
...
+        except requests.HTTPError:
+            LOGGER.exception(
+                "Unable to get container image grades %s - %s - %s",
+                container_grades_url,
+                resp.status_code,
+                resp.text,
+            )
+            raise
+        container_image = resp.json()
+        LOGGER.debug("Container grades for %s: %s", image_identifier, container_image)
```
mantomas, line 419: Unconditional log - can this be be logged outside of the loop? If my math is correct, this will be logged 120 times with current settings if Clair fail with grading for some reason.

operator-pipeline-images/operatorcert/pyxis.py
```diff
@@ -386,3 +386,52 @@ def wait_for_image_request(
...
+        if container_image.get("freshness_grades"):
+            LOGGER.info(
+                "Container grades for %s are completed: %s",
+                image_identifier,
+                container_image["freshness_grades"],
+            )
+            return container_image
+        LOGGER.info(
```
mantomas, line 427: Same concern as for the other log.

mantomas (approved):
Some minor comments, other than that LGTM 👍🏻

operator-pipeline-images/operatorcert/pyxis.py
Allda: The return value here is for consistency reasons as there are other functions in the module that have the same logic and return values.

operator-pipeline-images/operatorcert/pyxis.py
Allda: This was just a test log and I removed it.

operator-pipeline-images/operatorcert/pyxis.py
Allda: This is an intentional log that shows the progress of the watcher.

PR #588, opened 2024-02-07
Move package dependencies from optional section

Allda:
Dependencies marked as optional were not really optional and it is better to have them as a required dependencies.


haripate (approved):
LGTM

PR #592, opened 2024-02-19
Merge PR if no bundle is added and ci.yaml is updated

Allda:
Due to a Tekton limitation, we can't execute a merge task if any previous tasks have been skipped. This usually happens when no bundle is added and ci.yaml file is updated. Until Tekton enables this workflow I am temporarily moving the merge task into finally section.

The issue has been created in Tekton upstream to fix this issue https://github.com/tektoncd/pipeline/issues/7680

JIRA: ISV-4476

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/community-hosted-pipeline.yml
@@ -598,14 +598,18 @@ spec:
           subPath: preflight-trigger
           workspace: results
 
+  finally:
     # merge PR
     - name: merge-pr
       when:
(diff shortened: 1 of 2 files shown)


wcheang:
Are we sure this won't lead to unintended merging? If some of the previous tasks fail (e.g. validation or preflight), shouldn't the PR not be merged? But with this change all PR will be merged as long as the ci.yaml has `merge` set to `true`.

Allda:
> Are we sure this won't lead to unintended merging? If some of the previous tasks fail (e.g. validation or preflight), shouldn't the PR not be merged? But with this change all PR will be merged as long as the ci.yaml has `merge` set to `true`.

Of course you are right and than you for catching it!! I added one more condition to merge it only when pipeline passed.

wcheang (approved):

tomasbakk (approved):

PR #593, opened 2024-02-20
[ISV-4530] Fix static result comment rendering issue

mporrato:
`operator-sdk bundle validate` can return multi-line error messages that break the rendering of the result table in the comment added by the bot to the PR.

This patch just strips any line break from the error messages when building the markdown table.

Example comments: [without the fix](https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/3975#issuecomment-1924414561) and [with the fix](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/229#issuecomment-1954305913).


Allda (approved):

haripate (approved):
LGTM

PR #594, opened 2024-02-20
[ISV-4487,ISV-4488] isv pipelines use parse-repo-changes task

tomasbakk:
closes [ISV-4487](https://issues.redhat.com/browse/ISV-4487) and [ISV-4488](https://issues.redhat.com/browse/ISV-4488)


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -163,6 +164,56 @@ spec:
...
+      workspaces:
+        - name: pr
+          workspace: repository
+          subPath: src
+        - name: base
+          workspace: repository
+          subPath: base
+
     # Get the bundle path
     - name: get-bundle-path
```
Allda, line 218: This task still uses the path to a bundle from the title. You can replace it with output from detect changes and remove this task completely.

Allda: Thre is also task `bundle-path-validation` that can be removed since the same functionality is provided by detect changes task.

Allda (commented):
The changes present in this PR look fine but there are a few more tasks that still depend on the PR title and should be removed.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
tomasbakk: on it

tomasbakk:
@Allda I did the changes although I kept the `bundle-path-validation` itself intact as it is also used by the ci-pipeline. If I got the right impression of how ci-pipeline works, we are not able to substitute it with detect-changes task. But please correct me if I am wrong.

tomasbakk:
@Allda I did the changes although I kept the `bundle-path-validation` itself intact as it is also used by the ci-pipeline. If I got the right impression of how ci-pipeline works, we are not able to substitute it with detect-changes task. But please correct me if I am wrong.

Allda:
> `bundle-path-validation`

That's correct. We can keep the task there and just remove it from the hosted and release pipelines.

Allda (approved):

PR #595, opened 2024-02-21
Build fragment images from modified catalogs

Allda:
A new task that builds fragment images is added into hosted pipelines. The task generates a dockerfile, builds and pushes a fragments into registry.

JIRA: ISV-4492


operator-pipeline-images/Dockerfile
```diff
@@ -55,10 +56,9 @@ RUN curl -LO https://github.com/operator-framework/operator-registry/releases/do
     mv operator-sdk_linux_${ARCH} /usr/local/bin/operator-sdk
 
 RUN useradd -lms /bin/bash -u "${USER_UID}" user
-
 RUN useradd -lu "${PODMAN_USER_UID}" podman; \
-    echo podman:10000:5000 > /etc/subuid; \
-    echo podman:10000:5000 > /etc/subgid;
+    echo podman:10000:5000 >> /etc/subuid; \
+    echo podman:10000:5000 >> /etc/subgid;
```
mporrato, line 61: We can remove those since the version of `useradd` shipped with fedora 38 already generates subuid and subgid mappings when creating new users.

mporrato (commented):
I think you forgot to add the yaml file for the `build-fragment-images` task.

Allda:
> I think you forgot to add the yaml file for the `build-fragment-images` task.

Yes, I realized it today. The task is already in the PR.

operator-pipeline-images/Dockerfile
Allda: Fixed

mporrato (approved):

operator-pipeline-images/Dockerfile
```diff
@@ -54,11 +55,8 @@ RUN curl -LO https://github.com/operator-framework/operator-registry/releases/do
     chmod +x operator-sdk_linux_${ARCH} && \
     mv operator-sdk_linux_${ARCH} /usr/local/bin/operator-sdk
 
-RUN useradd -lms /bin/bash -u "${USER_UID}" user
-
-RUN useradd -lu "${PODMAN_USER_UID}" podman; \
-    echo podman:10000:5000 > /etc/subuid; \
-    echo podman:10000:5000 > /etc/subgid;
```
wcheang, line 61: Why are these lines being removed?

wcheang:
I just have one question about the 2 lines being removed from the Dockerfile. The rest looks fine to me.

operator-pipeline-images/Dockerfile
Allda: As Maurizio pointed out in the previous comment the subuid and subgid are automatically populated now https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/595#discussion_r1502610662

mporrato: This seems to work and it also gets rid of a warning we were getting before:

```
RUN useradd -lms /bin/bash -u "${USER_UID}" user ; \
    useradd -lu "${PODMAN_USER_UID}" podman; \
    setcap cap_setuid+ep /usr/bin/newuidmap ; \
    setcap cap_setgid+ep /usr/bin/newgidmap ; \
    echo -e "user:1:999\nuser:1001:5000" | tee /etc/subuid >/etc/subgid
```
At this point we can probably also get rid of the podman user.
If it wasn't itself based on fedora:latest I would suggest switching to https://quay.io/repository/containers/podman as the base image as there seem to be other [quirks](https://github.com/containers/podman/blob/main/contrib/podmanimage/stable/Containerfile) involved in getting podman working properly inside a container.

Allda: I had to revert the change to the previous implementation. This time both podman and buildah work for our purposes. Please review it again.

wcheang (approved):

PR #599, opened 2024-02-27
[ISV-4495] Tag bundle image with version

mporrato:


Allda (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/community-release-pipeline.yml
```diff
@@ -258,7 +258,7 @@ spec:
         - name: index_images
           value: "$(tasks.get-supported-versions.results.indices)"
         - name: bundle_pullspec
-          value: "$(params.registry)/$(params.image_namespace)/$(tasks.detect-changes.results.added_operator):$(tasks.detect-changes.results.added_bundle)-$(params.git_head_sha)"
```
Allda, line 261: This is a good fix! We often see release pipeline failures because of this issue with commi-specific tags.

PR #600, opened 2024-02-27
[ISV-4596] Single PR can update only one operator within catalogs

mantomas:
- validation logic moved from Bash to Python and refactored, so the fail message lists all detected issues in the fail (exception) message
- all conditionals also moved to Python script, Bash only reads the output from json

JIRA: ISV-4596

### DIFF
--- a/operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
+++ b/operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
@@ -1,4 +1,5 @@
 """Detect which operator bundles are affected by a PR"""
+
 import argparse
 import json
 import logging
@@ -6,7 +7,7 @@ import os
 import pathlib
 import re
 from dataclasses import dataclass, field
-from typing import Optional
+from typing import Optional, Any
 from operator_repo import Repo as OperatorRepo
 
 from github import Auth, Github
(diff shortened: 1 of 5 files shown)


operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
```diff
@@ -229,6 +230,80 @@ def union(self) -> set[AffectedCatalog]:
...
+    Raises a ValidationError if the result violates
+    any of the constraints.
+
+    Args:
+        result: Dictionary with the detected changes
+    """
+    message = ""
+
+    if len(extra_files := result.get("extra_files", [])) > 0:
+        message += f"The PR affects non-operator files: {sorted(extra_files)} "
```
Allda, line 252: Please update the error message with a nicer formatting. This way the error will contain one big line with no linebreaks. Either add a line break at the end of the message or append each message in a list and format it when raising ValidationError

mavaras:
Apart from Ales comment, LGTM.

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
```diff
@@ -497,7 +572,10 @@ def detect_changes(
         **catalog_operators,
         "extra_files": list(non_operator_files),
     }
-    return result
+
+    _validate_result(result)
```
Allda, line 576: It would be better to export these 2 functions from the detected changes. The validation and enrichment logic should be separated from the detect changes as it has very little in common.

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
mantomas: Good point, added line-breaks.

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
mantomas: Function execution moved to the main() and tests separated.

operator-pipeline-images/tests/entrypoints/test_detect_changed_operators.py
```diff
@@ -388,20 +423,22 @@ def test_detect_changes(
         Repo(before_dir),
         "https://example.com/foo/bar/pull/1",
     )
-
+    print(result)
```
Allda, line 426: Debug log?

Allda (approved):
One small comment otherwise it looks good.

operator-pipeline-images/tests/entrypoints/test_detect_changed_operators.py
mantomas: removed

PR #602, opened 2024-02-28
Unify and cleanup pipeline params

Allda:
Some pipeline params were no longer user and some had different names across multiple pipelines.

This commit removes unused params and rename existing to common names.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
@@ -226,19 +226,6 @@ spec:
           workspace: pipeline
           subPath: src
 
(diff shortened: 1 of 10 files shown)


mavaras (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
```diff
@@ -226,19 +226,6 @@ spec:
           workspace: pipeline
           subPath: src
 
-    - name: get-organization
```
tomasbakk, line 229: if we get rid of this task from the pipeline, are not going to be issues with subsequent tasks that are using it's value? for example `reserve-operator-name`

tomasbakk (approved):

PR #604, opened 2024-02-29
Skip index image import in community tests

Allda:
The index images are not needed in the community workflow and can be skipped.


mantomas (approved):
LGTM

PR #611, opened 2024-03-08
Replace verify-ci-results with evaluate-preflight-results task

Allda:
The isv hosted pipeline now uses a common task to evaluate preflight results that is shared between pipelines. This change brings us closer to a merge of isv and community pipelines.

JIRA: ISV-4632


jedinym (approved):
LGTM

PR #612, opened 2024-03-08
Fix a graph loop test

Allda:
The loop test needs to reset the visited list in case there are more than one path from the current node. Before this commit the update graph path could overlap which causes false negatives.

This fixes a false negatives in case there are multiple paths in upgrade graph.


mantomas (approved):
LGTM

PR #618, opened 2024-03-21
Add a migration script between old and new community repository

Allda:
A script copies images from the old quay namespace to a new one. In case a repo doesn't exists it is created and only a tag diff is copied to the destination to reduce time.

JIRA: ISV-4510


RichardPlesnik (approved):
Great job! LGTM :+1:

PR #622, opened 2024-03-25
[ISV-4678] Lock hosted pipeline execution

mantomas:
Copy acquire-lease logic from community hosted pipeline.

JIRA: ISV-4678

PR #624, opened 2024-03-25
[ISV-4672] release-lease task deletes lock not owned by the pipelinerun

mporrato:
Fix the issue by tagging leases with the pipelinerun name and removing all leases with the same tag at the end of the pipelinerun.


Allda (approved):
There is another open PR with changes to the isv hosted pipeline that uses the locks. Please sync with @mantomas on who will rebase and resolve conflicts.

https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/622

PR #625, opened 2024-03-26
[ISV-4624] Cert project check in operator-release-pipeline is optional

mantomas:


ansible/roles/operator-pipeline/templates/openshift/tasks/copy-image.yml
```diff
@@ -28,6 +30,13 @@ spec:
       script: |
         set -xe
 
+        if [ "$(params.cert_project_required)" == "true" ]; then
```
mporrato, line 33: Shouldn't this condition be reversed?
Also, since the cert_project itself has little to do with the image copy, I wonder if it would be better to check the dest_image_registry_namespace_certproject parameter instead since (I suppose) it will be given an empty value when the pipeline is called from the community pipeline repo's merge webhook.

mantomas: Thanks, good catch - after some discussion I decided to not use the cert_project_required flag and instead split the task params. The check now uses the vendor_label (empty for community operators).

mporrato (approved):

jedinym (approved):
lgtm

PR #629, opened 2024-04-03
Skip preflight test only when previous attempt was sucesfull

Allda:
There was a bug introduced in recent commit that cause a preflight to be skipped even for previous attempts that weren't successful.

This commit fixes the workflow to skip a preflight only when test was successful and always execute preflight when previous attempt wasn't found or attempt failed.

JIRA: ISV-4741

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
@@ -819,6 +819,11 @@ spec:
           value: "$(params.prow_kubeconfig_secret_key)"
         - name: test_result_path
           value: "$(tasks.get-ci-results-attempt.results.preflight_results)"
+        # Skip preflight trigger in case there was a previous SUCCESSFUL test
+        # The test could be executed in CI or hosted pipeline but it needs
+        # to be successful. Previous failed test don't skip this task.
+        - name: skip
+          value: "$(tasks.get-ci-results-attempt.results.preflight_results_exists)"
         - name: github_token_secret_name
           value: "$(params.github_token_secret_name)"
         - name: github_token_secret_key
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml
@@ -55,6 +55,11 @@ spec:
     - name: test_result_path
       description: "A path to existing test results. This indicates if we can skip a preflight test."
       default: "none"
+
+    - name: skip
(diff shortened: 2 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -819,6 +819,11 @@ spec:
           value: "$(params.prow_kubeconfig_secret_key)"
         - name: test_result_path
           value: "$(tasks.get-ci-results-attempt.results.preflight_results)"
+        # Skip preflight trigger in case there was a previous SUCCESSFUL test
+        # The test could be executed in CI or hosted pipeline but it needs
+        # to be successful. Previous failed test don't skip this task.
+        - name: skip
+          value: "$(tasks.get-ci-results-attempt.results.preflight_results_exists)"
```
johnbe11, line 826: @Allda the assignment of this var in get-ci-results-attempt.yml sets it to the file path which contains the results, but I'm not seeing where we assess the content of that file as opposed to the filepath itself?

johnbe11 (changes requested):
Overall looks good, with one question about how that file/content variable is being evaluated.

mavaras (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
Allda: Can you give me a line where you see that? The variable is populated with a result status from what I see.
```
CHECK_RESULTS=$(cat $RESULTS_FILE | jq -r '.passed')
echo -n "$CHECK_RESULTS" | tee $(results.preflight_results_exists.path)
```

johnbe11: Right, I'm seeing the results written to the file referenced by that variable (results.preflight_results_exists.path) happening here:
https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/bf23c3abb7ec6b29e775abb7f866c0825db5a6d9/ansible/roles/operator-pipeline/templates/openshift/tasks/get-ci-results-attempt.yml#L89

Maybe I'm misreading the evaluation here?  It looks like we're then just reading the filepath string itself, instead of the content of the file referenced by the filepath:
       `value: "$(tasks.get-ci-results-attempt.results.preflight_results_exists)"`

johnbe11: I think I misread that evaluation.  This looks good to go.

johnbe11 (approved):

PR #633, opened 2024-04-10
[ISV-4299] Add explicit token permissions to release step

mavaras:
The cause of the "_Resource not accessible by integration_" error seems to be something related to the github token permissions:
* https://github.com/mathieudutour/github-tag-action/issues/138
* https://stackoverflow.com/questions/75995802/resource-not-accessible-by-integration-github-action-fails-for-pushed-commit

I'd be worth to try add this explicit declaration of token permissions to the _release_ job and observe if the problem does not appear anymore.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -133,6 +133,7 @@ jobs:
   release:
     name: Github release
     runs-on: ubuntu-latest
+    permissions: write-all
     needs:
       - deploy-prod
     steps:


johnbe11 (approved):
@mavaras this seems like a good way to verify whether the token permissions are the blocker.

For the longer term, I wonder if having "write-all" there might be overly broad? Meaning, if this works, it might be worthwhile to then clarify which particular write permission (e.g. contents vs issues) is the specific culprit.

mavaras:
Testing pme-bot: #ISV-4299

PR #642, opened 2024-04-30
Move static check from common to isv

Allda:
The checks are usually skipped by support team so it doesn't make sense to have them in isv pipelines.


mporrato (approved):
Apart from the misleading commit message, LGTM.
I believe ISV-4831 together with ISV-4806 should resolve most, if not all, the issues we are seeing in the ISV pipeline, so perhaps we may want to wait before merging this.
I've started doing some work on ISV-4831 [here](https://github.com/mporrato/operator-repo/compare/main...relax-update-graph-checks).

Allda:
> Apart from the misleading commit message, LGTM. I believe ISV-4831 together with ISV-4806 should resolve most, if not all, the issues we are seeing in the ISV pipeline, so perhaps we may want to wait before merging this. I've started doing some work on ISV-4831 [here](https://github.com/mporrato/operator-repo/compare/main...relax-update-graph-checks).

@mporrato The cert ops team is right about adding tests into isv pipeline without documenting it first in the official doc. Until we have it documented (it is currently part of FBC epic) I want to move it out of the isv space. We may reconsider adding it back after a discussion with cert ops.

PR #643, opened 2024-04-30
updating all scripts to exit on non-zero status

acornett21:


mporrato (approved):
LGTM, but I think you missed a couple of tasks:
- verify-whitelisted-index-image.yml
- preflight-trigger.yml

acornett21:
@mporrato good catch on the `verify-whitelisted-index-image` task, we already take care of this in `preflight-trigger` but in a more inline way where it is needed... See below as an example:

https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/preflight-trigger.yml#L103

Allda (approved):

PR #645, opened 2024-05-03
[ISV-4806] - refactor check_replaces_availability test

tomasbakk:
I spent some time trying to find the most convenient way how to check for the previous replaced bundle but in the end, this seemed to me like the most simple way how to get rid of the KeyErrors.
Since naming of the previous versions is either specified by "v" prefix+actual bundle version or just a numeric value of the bundle, I added additional check that is looking into the operator folder for a replaces bundle with and w/o the prefix.


mporrato (changes requested):
This won't work for operators where the bundle directory names are more complex, for example https://github.com/redhat-openshift-ecosystem/certified-operators/pull/3722 .

I would instead build a dictionary to map the versions from the csv to the directory names and use it to lookup the directory name when instantiating the replaced bundle with something like this:

```python
ver_to_dir = {x.operator_version: x.csv_operator_version for x in bundle.operator.all_bundles()}
replaces_bundle = bundle.operator.bundle(ver_to_dir[replaces_version])
```

This should work in all situations.

tomasbakk:
> This won't work for operators where the bundle directory names are more complex, for example [redhat-openshift-ecosystem/certified-operators#3722](https://github.com/redhat-openshift-ecosystem/certified-operators/pull/3722) .
> 
> I would instead build a dictionary to map the versions from the csv to the directory names and use it to lookup the directory name when instantiating the replaced bundle with something like this:
> 
> ```python
> ver_to_dir = {x.csv_operator_version: x.operator_version for x in bundle.operator.all_bundles()}
> replaces_bundle = bundle.operator.bundle(ver_to_dir[replaces_version])
> ```
> 
> This should work in all situations.

Oh, yes, you are right. I forgot that there is yet another option how to annotate the bundle.  Your solution works for all bundles I had tested so I switched them.

tomasbakk:
> What are the changes in pdm.lock? Looks like you are using an older version of pdm that uses a different structure of the lock file and it makes it difficult to spot the differences. Can you please upgrade your pdm and do a refresh of the lock file?

I updated the pdm and refreshed the lock and now it looks good.

mporrato (approved):

PR #654, opened 2024-05-14
updating preflight sha for version 1.9.5

acornett21:
```
❯ crane digest quay.io/redhat-isv/preflight-test:1.9.4
sha256:138709be6c93714223077341f29c114e2c6d8769e895df1434a26de6c663c7e5

~ 
❯ crane digest quay.io/redhat-isv/preflight-test:1.9.5
sha256:8cd846a504d869a79186b508c3b5e973eda6b1fa67c5095c4319cdc03d332f88
```

PR #653, opened 2024-05-14
Add FBC Makefile + doc

Allda:
The Makefile will be used by operator owners to generate catalogs from templates. This file simplifies the workflow and hides any opm related commands from a user.

The commit also introduces a user documentation for a FBC workflow.

JIRA: ISV-4512

PR #655, opened 2024-05-14
updating pdm lock file with latest dependencies

acornett21:


acornett21:
@Allda I have no idea what I'm doing with python here, but I'd like to be able to unblock

- #654

Allda:
@acornett21 We already have that covered in https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/653 I'll merge your PR with preflight version once the PR with dependencies is reviewed. I assume it is going to be by the end of today.

PR #663, opened 2024-05-29
[ISV-4915] Fix corner cases in check_operator_name

mporrato:
Inconsistencies between operator names defined within a bundle are now fatal only if the bundle being added is the first bundle of a new operator.
Inconsistencies across bundles still cause failures only if the inconsistency is caused by the bundle being added and just result in a warning if the inconsistency was pre-existing.


mantomas (approved):
LGTM

tomasbakk (approved):

PR #665, opened 2024-06-05
Bundle and catalogs are mutually exclusive

Allda:
The pipeline refuse RPs with mix of bundle and catalog changes.

JIRA: ISV-4639

### DIFF
--- a/operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
+++ b/operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
@@ -260,6 +260,14 @@ def _validate_result(result: dict[str, list[str]]) -> None:
         message += f"The PR deletes existing bundles: {sorted(deleted_bundles)}\n"
     if len(added_bundles := result.get("added_bundles", [])) > 1:
         message += f"The PR affects more than one bundle: {sorted(added_bundles)}\n"
+    if len(
+        affected_catalog_operators := result.get("affected_catalog_operators", [])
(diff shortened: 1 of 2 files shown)


operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
```diff
@@ -260,6 +260,14 @@ def _validate_result(result: dict[str, list[str]]) -> None:
         message += f"The PR deletes existing bundles: {sorted(deleted_bundles)}\n"
     if len(added_bundles := result.get("added_bundles", [])) > 1:
         message += f"The PR affects more than one bundle: {sorted(added_bundles)}\n"
+    if len(
+        affected_catalog_operators := result.get("affected_catalog_operators", [])
+    ) > 0 and (len(affected_operators) > 0):
```
mporrato, line 265: If we do it like this:

```suggestion
    if len(
        affected_catalog_operators := result.get("affected_catalog_operators", [])
    ) > 0 and (len(result.get("affected_bundles", [])) > 0):
```

we could allow changing ci.yaml in the same PR, avoiding an extra PR for something that does not require any extra processing on our side.

mporrato (approved):

operator-pipeline-images/operatorcert/entrypoints/detect_changed_operators.py
Allda: Makes sense. Updated

PR #670, opened 2024-06-14
[ISV-4972] Fix community integration test failure

mporrato:
This avoids introducing a double dash in the tag of the index reference copied from IIB which would cause a mismatch with the reference passed to preflight.


mantomas (approved):
LGTM 👍🏻

PR #678, opened 2024-06-25
Fix race condition in IIB build step

Allda:
There was an race condition in the " | tee ..." command that sometime cause the output file was empty. This caused a unpredictable issue later in a pipeline.


mantomas (approved):
LGTM

PR #683, opened 2024-06-28
[ISV-4837] Incorrect behaviour of the operator pipelines when a PR affects only ci.yaml. 

haripate:
This PR is initiated to resolve an issue of the operator CI pipeline disruption by PR #667.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
@@ -786,7 +786,7 @@ spec:
     - name: make-bundle-repo-public
       runAfter:
         - build-bundle
-      when: 
+      when:
         - *operatorOrBundleChange
       taskRef:
         name: set-quay-repo-visibility
@@ -853,7 +853,7 @@ spec:
     - name: add-bundle-to-index
       runAfter:
         - add-fbc-fragments-to-index
-      when: 
+      when:
         - *operatorOrBundleChange
       taskRef:
         name: add-bundle-to-index
@@ -888,7 +888,7 @@ spec:
     - name: make-index-repo-public
       runAfter:
         - add-bundle-to-index
-      when: 
+      when:
         - *operatorOrBundleChange
       taskRef:
         name: set-quay-repo-visibility
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
@@ -1207,7 +1207,7 @@ spec:
     - name: upload-pipeline-logs
       taskRef:
         name: upload-pipeline-logs
-      when: 
+      when:
         - *certProjectExists
         - *bundleAdded
       params:
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
@@ -489,8 +489,8 @@ spec:
       params:
         - name: pipeline_image
           value: "$(params.pipeline_image)"
-        - name: operator_path
-          value: "$(tasks.detect-changes.results.operator_path)"
+        - name: bundle_path
+          value: "$(tasks.detect-changes.results.bundle_path)"
         - name: cert_project_required
           value: "$(params.cert_project_required)"
       workspaces:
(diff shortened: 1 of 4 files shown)


mporrato (approved):

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -1201,7 +1217,9 @@ spec:
     - name: upload-pipeline-logs
       taskRef:
         name: upload-pipeline-logs
-      when: *certProjectExists
+      when:
+        - *certProjectExists
+        - *bundleAdded
```
Allda, line 1222: I think you're putting a `bundleAdded` condition in places where it doesn't belong. Why do we need to skip `upload-pipeline-logs` if a bundle is absent?

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -237,6 +237,11 @@ spec:
     - name: content-hash
       runAfter:
         - read-config
+      when:
```
Allda, line 240: This can't be skipped using a when because the result of this task is used in other places.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -1056,6 +1070,8 @@ spec:
         - evaluate-preflight-result
       taskRef:
         name: get-ci-results
+      when:
```
Allda, line 1073: This can't be skipped using a when because the result of this task is used in other places.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -983,6 +998,7 @@ spec:
         - input: "$(tasks.detect-changes.results.affected_catalogs)"
           operator: in
           values: [""]
+        - *bundleAdded
```
Allda, line 1001: The task already skips a upload process inside of the task. We don't have to skip it here.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -898,6 +911,8 @@ spec:
     - name: get-ci-results-attempt
       runAfter:
         - make-index-repo-public
+      when:
```
Allda, line 914: This can't be skipped using a when because the result of this task is used in other places.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -465,6 +466,8 @@ spec:
         name: content-hash
       runAfter:
         - apply-test-waivers
+      when:
```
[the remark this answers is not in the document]
haripate: I am not sure if this is the expected behavior for new FBC-supported operator pipelines but here is the tested pipeline run: https://console-openshift-console.apps.pipelines-stage.0ce8.p1.openshiftapps.com/k8s/ns/haripate-playground/tekton.dev~v1~PipelineRun/operator-hosted-pipeline-runvf5rh

The test PR: https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/266

If this is not correct behavior then I would suggest having the If condition in a script of each affected tekton task rather than the when condition.

Allda:
@haripate Based on the original description of the ticket I think the original bug is already fixed in the following PRs:

https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/697
https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/696

Once these PRs are merged can you verify it is no longer reproducible?

haripate:
> @haripate Based on the original description of the ticket I think the original bug is already fixed in the following PRs:
> 
> #697 #696
> 
> Once these PRs are merged can you verify it is no longer reproducible?

Ack. Sure, I will verify the results.

RichardPlesnik:
> > @haripate Based on the original description of the ticket I think the original bug is already fixed in the following PRs:
> > #697 #696
> > Once these PRs are merged can you verify it is no longer reproducible?
> 
> Ack. Sure, I will verify the results.

I am currently running the community pipeline while working on ISV-5013 and in the current `main` branch, there is still issue with `yaml-lint` step .

There is a missing condition for `tasks.detect-changes.results.bundle_path` which leads to `[Errno 2] No such file or directory: '' ` if no bundle is changed because yamllint is receiving empty string as an parameter.

haripate:
The issue mentioned in this PR is resolved.

PR #684, opened 2024-06-28
Check permission script supports user's repositories

Allda:
When user forks a repository under its own namespace the Github API raises exception since the organization doesn't exists and instead user namespace is used. This commits add a support of such a repository to allow running pipelines from forks.


haripate (approved):
LGTM

PR #686, opened 2024-07-01
[ISV-5002] Remove CA data from kubeconfig files

mporrato:
The certificate-authority-data field in ci-pipeline-kubeconfig contained outdated intermediate CA cert info and was therefore removed. Since the API endpoint uses a public CA, there is no need to include its certs in the kubeconfig file.

Additionally:
- Removed empty certificate-authority-data from prow kubeconfigs
- Dropped kubeconfigs for retired RHC4TP clusters


Allda (approved):
There is one more place where the removed secret is still referenced. Please remove it as well. Otherwise :+1: 
https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/docs/pipeline-env-setup.md#ocp-registry-kubeconfig

RichardPlesnik (approved):

PR #689, opened 2024-07-12
[ISV-5072] Ignore malformed operator names in old bundles

mporrato:
This fixes the issue with the 2.0.1 bundle of the certified anzo-operator causing a validation failure in newer bundles.

I've tested the updated version of the common test suite against the entire certified-operators repo and no other operators are now reporting failures caused by older bundles (except, of course, for operators containing a single bundle).


johnbe11 (approved):
LGTM, thanks for the quick fix Maurizio!

PR #692, opened 2024-07-19
Upgrade pdm lock version

mporrato:
Upgrade the pdm lock file structure to the latest version. This solves the issue with the safety test which is causing the github validation workflow to fail, which prevents the container image from being built, ultimately blocking the integration tests.


johnbe11 (approved):
LGTM

mantomas (approved):

PR #694, opened 2024-07-29
[ISV-4964] Implement Preflight version invalidation tool

jedinym:
https://issues.redhat.com/browse/ISV-4964

These changes have been tested in dry run mode in both stage and prod. The full test in stage is blocked by the Pyxis proxy release. After this PR is merged and tested in stage, I will manually deploy the CronJob to the prod cluster, run it manually and check that all is working properly.


operator-pipeline-images/operatorcert/entrypoints/update_preflight_versions.py
```diff
@@ -0,0 +1,182 @@
+"""
```
wcheang, line 1: disable_preflight_versions or invalidate_preflight_versions would probably be a more precise name for the script? Update is a bit too general.

wcheang (approved):
A couple of minor comments, but the code mostly looks good.

ansible/roles/preflight-invalidation/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+preflight_invalidation_namespace: preflight-invalidation
+
+preflight_invalidation_private_key_local_path: ../../vaults/{{ env }}/operator-pipeline.key
+preflight_invalidation_private_cert_local_path: ../../vaults/{{ env }}/operator-pipeline.pem
+
+preflight_invalidation_image_pull_spec: quay.io/redhat-isv/operator-pipelines-images:latest
```
mporrato, line 7: We probably want to use the `released` tag here rather than `latest`, at least for prod.
For development, we can always override it at run time when running the playbook.

mporrato (changes requested):

mporrato:
Are you planning to add the new role in the main deployment playbook? I think this should be part of the normal release workflow.

mporrato (approved):
Sorry for the delay.
LGTM :+1:

PR #696, opened 2024-07-30
[ISV-4829] Release pipeline skips catalog build and distribution for FBC

mporrato:


ansible/roles/operator-pipeline/templates/openshift/tasks/resolve-pr-type.yml
```diff
@@ -0,0 +1,48 @@
...
+        set -xe
+
+        if [ "$(params.fbc-enabled)" == "true" ] ; then
+          # PRs for fbc operators can result in the build of bundles only, catalogs only, both or none
+          if [ -n "$(params.bundle_path)" ] ; then
+            echo -n true >"$(results.add-bundle.path)"
+          else
+            echo -n false >"$(results.add-bundle.path)"
+          fi
+          if [ -n "$(params.affected_catalogs)" ] || [ -n "$(params.affected_catalogs)" ] ; then
```
Allda, line 33:
```suggestion
          if [ -n "$(params.affected_catalogs)" ] || [ -n "$(params.affected_catalog_operators)" ] ; then
```

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -236,12 +236,33 @@ spec:
...
+        - read-config
+      params:
+        - name: bundle_path
+          value: "$(tasks.detect-changes.results.bundle_path)"
+        - name: affected_catalogs
+          value: "$(tasks.detect-changes.results.affected_catalogs)"
+        - name: affected_catalog_operators
+          value: "$(tasks.detect-changes.results.affected_catalog_operators)"
+        - name: fbc-enabled
+          value: "$(tasks.read-config.results.fbc-enabled)"
```
Allda, line 253: It looks like `read-config` can't handle PRs with catalog changes only. It only depends on bundle path that is empty. We will need to update the `read-config` task to also accept `affected_catalog_operators` result.

We can do it as part of this PR or create a separate ticket for it.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -236,12 +236,33 @@ spec:
...
+
     # Compute a hash of the content so pipeline logs can be uploaded with it
     - name: content-hash
       runAfter:
-        - read-config
+        - resolve-pr-type
       taskRef:
         name: content-hash
+      when:
```
Allda, line 261: Some tasks depend on a result of this one. In case we skip it using `when` the follow-up task will be skipped.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
Allda: I created https://issues.redhat.com/browse/ISV-5115 to fix this.

ansible/roles/operator-pipeline/templates/openshift/tasks/resolve-pr-type.yml
mporrato: Thanks for spotting that, fixed.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
mporrato: Not sure what you mean: read-config just reads out info from ci.yaml and it locates the file based on operator_path which should be set even when the change only affects the catalog. From the tests I've done so far it seems to be working just fine.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
mporrato: Thanks, fixed.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
Allda: Based on my testing when a user submits just a catalog change the detect changes task doesn't return a value that is later consumed by the read-config tasks and read-config then returns default values (fbc = false).

You can see it here: https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/1627

I am going to take care of it in the ticket I linked above.

PR #697, opened 2024-07-31
Integration tests for FBC workflow

Allda:
Tests cover a new FBC workflow. There are 2 new tests and bunch of refactoring in the Ansible.

2 new tests covers:
 - fbc bundle submission
 - fbc catalog submission

Tests are now controlled by dedicated ansible inventory and all test data are customized using Ansible role.

JIRA: ISV-4498


ansible/roles/integration_tests/tasks/test_data.yml
```diff
@@ -2,33 +2,96 @@
...
+    FBC_CATALOG="{{ integration_tests_fbc_catalog }}"
+
+    echo "FBC catalog is: $FBC_CATALOG"
+
+    if [[ $FBC_CATALOG == "True" ]]; then
+      # In case of FBC catalog submission a bundle needs to be pushed in the target branch
+      BUNDLE_BRANCH=$UPSTREAM_BRANCH
+      git checkout $BUNDLE_BRANCH
+    else
+      git checkout -b $BUNDLE_BRANCH $UPSTREAM_BRANCH
```
BorekZnovustvoritel, line 71: This command could break if there were some mistakes in the `BUNDLE_BRANCH` or `UPSTREAM_BRANCH` variables. If they contained a space for any reason, this would be interpreted as multiple words/arguments instead of one. Best practice is to double-quote shell variable expansion.  This same problem occurs at more places in this file.
```suggestion
      git checkout -b "$BUNDLE_BRANCH" "$UPSTREAM_BRANCH"
```

BorekZnovustvoritel (approved):
Great initiative to reformat the code in a meaningful way! I have 2 rather minor comments related to this, but hey are mostly related to formatting.

mantomas (approved):
Very nice refactoring along with the new tests 👍🏻

ansible/roles/integration_tests/tasks/test_data.yml
Allda: Fixed

PR #700, opened 2024-08-05
Fix: read-config support FBC catalog submission

Allda:
The read-config didn't support a case where a FBC catalog only was submitted. This commit make a task compatible with FBC by finding a ci.yaml file in the operator directory based on which catalog operator was submitted.

JIRA: ISV-5115


tomasbakk:
LGTM :+1:

tomasbakk (approved):

PR #701, opened 2024-08-06
Tests: reduce index count in tests

Allda:
The bottleneck for integration tests is IIB which builds index images in series. The previous configuration always released 2 catalogs per tests. This caused a significant delays and sometime IIB builds failed with timeout due to a long queue.

This commit reduce amount of affected index images to only one and at the same time makes isv-fbc-catalog test mandatory. All bugs that cause a test to fails have been already fixed and there is no need for making this test optional anymore.


mantomas (approved):
LGTM

PR #703, opened 2024-08-08
Fix certification-project-check arguments

mporrato:


ansible/roles/operator-pipeline/templates/openshift/tasks/cert-project-check.yml
```diff
@@ -6,8 +6,8 @@ metadata:
 spec:
   params:
     - name: pipeline_image
-    - name: bundle_path
-      description: path indicating the location of the certified bundle within the repository
+    - name: operator_path
```
Allda, line 9: Hosted and release pipelines still use `bundle_path` argument.

Allda (approved):

PR #704, opened 2024-08-08
Allow catalog operator removal action

Allda:
With this commit operator owners are able to remove an operator from a catalog. The removal process is handled by the IIB endpoint and is only supported for FBC operators.

JIRA: ISV-5139


Allda:
There are a bunch of examples where I tested this feature. You can check that for the review process https://github.com/Allda/community-operators-pipeline-preprod/pulls?q=is%3Apr+is%3Aclosed

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
```diff
@@ -845,24 +845,26 @@ spec:
           workspace: repository
           subPath: src
 
-    - name: add-fbc-fragments-to-index
+    - name: build-fbc-index-images
       runAfter:
         # This is just a temporary location until we have a proper
         # FCB branch based on config
```
mantomas, line 851: Not related to the logic, but there are multiple comments accross files with this "temporary location" - if there are no further plans to move the tasks, shouldn't this be removed?

mantomas (approved):
One partially related question but other than that LGTM. Pipeline output looks good.

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
Allda: I removed these comments.

tomasbakk (approved):

PR #710, opened 2024-08-14
Fix: Render only catalogs supported by an operator

Allda:
The fbc onboarding script is updated to only generate a template and catalog for ocp versions where the catalog exists. The script skips version where operator has never been shipped.

JIRA: ISV-5131


RichardPlesnik (approved):
LGTM :+1:

PR #712, opened 2024-08-19
Fix: bump opm version to produce compatible catalogs

Allda:
An older version of opm produced a catalog that OCP server couldn't read and skipped operators. This change updates opm and few other dependencies to fix the issue.


mantomas (approved):
LGTM

PR #718, opened 2024-08-26
Fix: Recognize operator bundle directory

Allda:
A detect changes task wasn't able to recognize bundle directories in the operator directory. It marked all directories as a bundle no matter its content. This caused issue when introducing a catalog-templates directory which is operator sub-directory but not a bundle.

This commit makes a parser more clever and uses a OperatorRepo probes to detect operator bundle dir. In case it is not recognized as a bundle it is automatically marked as "other" files owner by the operator.

This commit also simplifies a unit tests to reduce amount of fixtures objects size.

JIRA: ISV-5168


mporrato (approved):
Nice an simple solution and great simplification of the tests. :+1:

PR #722, opened 2024-09-05
[ISV-5194] Operator pipeline merge-pr parse the gh tool response

RichardPlesnik:
If `merge-pr` step of hosted pipeline fails, error message is included in the PR comment.

If merge fails because PR is marked as a draft or because there are new commits, pipeline passes without merge. Otherwise, pipeline fails.

Passing pipeline:
https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/1757

Merge failed because of draft:
https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/1760

Merge failed because of new commit:
https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/1759

Merge failed with different non-allowed error:
https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/1761

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/github-pipelinerun-summary.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/github-pipelinerun-summary.yml
@@ -164,6 +164,13 @@ spec:
           fi
         fi
 
+
+        if [[ -f "$(workspaces.output.path)/merge_error.txt" ]]; then
+          MERGE_ERROR=$(cat "$(workspaces.output.path)/merge_error.txt")
+          echo -e "\n## ❌ Not merged\n" >> $PR_NAME/comment.md
+          echo -e "PR cannot be merged for the following reasons: \`$MERGE_ERROR\`" >> $PR_NAME/comment.md
+        fi
+
         echo -e "\n## Troubleshooting\n\nPlease refer to the [troubleshooting guide]($DOC_LINK)." >> $PR_NAME/comment.md
         echo -e "\nRun \`/pipeline restart $(params.pipeline_name)\` in case of pipeline failure to restart a pipeline." >> $PR_NAME/comment.md
 
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/merge-pr.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/merge-pr.yml
@@ -21,6 +21,8 @@ spec:
       default: token
   workspaces:
     - name: source
+    - name: output
+      description: Scratch space and storage for the comment and related data
   results:
     - name: bool_merge
     - name: pr_merged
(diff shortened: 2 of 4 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/merge-pr.yml
```diff
@@ -84,21 +80,27 @@ spec:
...
         else
             echo "Cannot merge PR"
             echo -n "false" > "$(results.pr_merged.path)"
-            exit 1
+            # Pass details about merge error to summary comment task
+            echo -n $MERGE_ERROR | tee "$(workspaces.output.path)/merge_error.txt"
+            # Fail pipeline if the error is not explicitly allowed.
+            if [[ "$MERGE_ERROR" != "Pull Request is still a draft (mergePullRequest)" && "$MERGE_ERROR" != "Head branch was modified. Review and try the merge again. (mergePullRequest)" ]] ; then
+              exit 1
+            fi
```
mporrato, line 105: The message "Head branch was modified. Review and try the merge again. (mergePullRequest)" may be a bit too obscure for the average end user. Since we are handling it as a special case anyway, I would replace the message shown to the user with something more clear like "The PR was not merged because another pipeline is already in progress: please wait for the result of the new tests" or something like that.

mporrato (approved):

ansible/roles/operator-pipeline/templates/openshift/tasks/github-pipelinerun-summary.yml
```diff
@@ -164,6 +164,13 @@ spec:
           fi
         fi
 
+
+        if [[ -f "$(workspaces.output.path)/merge_error.txt" ]]; then
+          MERGE_ERROR=$(cat "$(workspaces.output.path)/merge_error.txt")
+          echo -e "\n## ❌ Not merged\n" >> $PR_NAME/comment.md
+          echo -e "PR cannot be merged for the following reasons: \`$MERGE_ERROR\`" >> $PR_NAME/comment.md
```
Allda, line 171: One of the message from the example you liked shows following message:

`PR cannot be merged for the following reasons: Pull request Pull request is closed (enablePullRequestAutoMerge)`

Does the double "Pull request" come directly from github?

Allda (approved):
It looks fine but please update the message Maurizio pointed out to make it more clear to users.

ansible/roles/operator-pipeline/templates/openshift/tasks/github-pipelinerun-summary.yml
RichardPlesnik: Yep, I've checked it out and it seems like there is multiple examples of error message where `Pull request` is duplicate. I can search for these cases and parse them, if we don't mind the difference between "official" error message.

PR #726, opened 2024-09-10
[ISV-4941] Upgrade to Fedora 40 and Python 3.12

RichardPlesnik:


BorekZnovustvoritel (approved):
LGTM :+1:

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.black]
-target_version = ['py311']
```
mporrato, line 2: Have you forgotten to refresh/commit the lock file? I would expect a change like this to cause a bunch of updates to the lock file.

mporrato: Just realized my comment wasn't clear enough: I wasn't referring to that specific line. Let me rephrase it:

Have you refreshed the lock file from a system with python 3.12?

operator-pipeline-images/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.fedoraproject.org/fedora:38
+FROM images.paas.redhat.com/exd-sp-guild-isv/base-images:f40-latest
```
Allda, line 1: This is an internal instance of the registry that is not available outside of the Red Hat VPN and you can't use it in public GitHub projects.

pyproject.toml
RichardPlesnik: I tried to install it in python3.12 venv  both locally and in integration tests and it was running fine with the current lock file. But I can regenerate the new one if you want.

operator-pipeline-images/Dockerfile
RichardPlesnik: Thanks for pointing it out. I tried it without the kerberos authentication but apparently forgot to turn of my VPN before that. I will change it to the public fedora image.

mporrato (approved):

Allda (approved):

PR #729, opened 2024-09-12
Run integration tests on PR

Allda:
A tests will be executed automatically on each PR after a PR gets approved.

To reduce amount of test run we don't run a test on PR push event and instead trigger it by label.

JIRA: ISV-4199


mavaras (approved):

tomasbakk:
lgtm

tomasbakk (approved):

PR #734, opened 2024-09-25
[RELEASE-1117] Add retry when requesting RADAS signature

mmalina:
We run the signing pipeline in Konflux a lot and users report that oftentimes one or a few of the requests fail.

In theory some retry should work on the Python library level, but it doesn't really seem to work. So this should help. See Jira for more details.

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/request-signature.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/request-signature.yml
@@ -62,19 +62,35 @@ spec:
         #! /usr/bin/env bash
         set -xe
 
+        MAX_RETRIES=3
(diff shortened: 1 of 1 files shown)


mmalina:
@Allda [the Build and push image test](https://github.com/redhat-openshift-ecosystem/operator-pipelines/actions/runs/11031869081/job/30641878474) failed. Can you check? It failed on docker auth. I know that normally PR events don't have access to secrets, so I'm not sure how this is supposed to work. Anything I need to do?

Allda (approved):

Allda:
@mmalina We haven't figured out how to support actions from an outside collaborator. Since this is not affecting a build part I'll merge it and build and the test will be executed right after that.

mmalina:
@Allda cool, thanks!

PR #742, opened 2024-10-14
[ISV-5336] Automation for pipeline re-run

mporrato:


Allda (approved):

mavaras (approved):
really good :+1:

mporrato:
I've updated the workflows to be compatible with github's ubuntu-24.04 image which was rolled out as the new ubuntu-latest in the middle of my tests and was rolled back to 22.04 shortly after. I've made sure the changes are still compatible with 22.04 so I'm going to keep them as part of this PR.

PR #750, opened 2024-11-27
[ISV-5341] fix broken link in the pipeline summary

tomasbakk:
This PR fixes the link in operator pipelines repo. Link redirects users to the 1st point of our documentation, from which they can move on to how the bundle should be created. Seemed to me like a good start since we do not have similar troubleshooting guide as we have for certified/marketplace operators. 
Actual point of reference is then updated in the PR template in 2nd [PR](https://github.com/redhat-openshift-ecosystem/community-operators-prod/pull/5557) that also fixes non-existent links in community-pipeline-prod repo.

---
Closes [ISV-5341](https://issues.redhat.com/browse/ISV-5341)

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/github-pipelinerun-summary.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/github-pipelinerun-summary.yml
@@ -128,7 +128,7 @@ spec:
 
         if [[ "$ORGANIZATION" == "community-operators" ]]
         then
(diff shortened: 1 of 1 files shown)


tomasbakk:
> The new link doesn't really lead to a troubleshooting document but I guess it will do for now as I don't think we have anything better. We should probably open a Jira to create a simple doc showing how to fix the most common issues.

Yeah, I looked for a something better but there is nothing similar to certified/marketplace. I will create a jira to write something up.

PR #753, opened 2024-12-06
[ISV-5446] static check for catalog bundle images

mantomas:
This PR adds logic to validate file based catalogs in static checks, with first fbc check `check_bundle_images_in_fbc`.

Example pipeline run [can be found here](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/315). The validation is done thanks to config.yaml key [allowed_bundle_registries](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/blob/f834b495bfa2ecb41bd656addad1947c414c10c3/config.yaml#L17). In the test run two catalogs were updated, but only one failure is detected, as one of the bundles use allowed registry.

If the key `allowed_bundle_registries` is missing in config.yaml, then the  bundle image validation is internally skipped (without the config the new check is turned off).

Closes ISV-5446

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/run-static-tests.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/run-static-tests.yml
@@ -19,6 +19,9 @@ spec:
     - name: bundle_version
       description: Version of the bundle under test.
 
+    - name: affected_catalog_operators
+      description: All catalog operators affected with the change.
+
     - name: github_token_secret_key
       description: The key within the Kubernetes Secret that contains the GitHub token.
       default: token
@@ -60,8 +63,8 @@ spec:
         #!/usr/bin/env bash
         set -xe
 
-        if [ -z "$(params.bundle_path)" ]; then
-          echo "No bundle added or changed, skipping tests"
+        if [ -z "$(params.bundle_path)" ] && [ -z "$(params.affected_catalog_operators)" ]; then
+          echo "No bundle or catalog added/changed, skipping tests"
           echo -n "" | tee $(results.messages_count.path)
           echo -n "" | tee $(results.failures_count.path)
           exit 0
@@ -86,7 +89,8 @@ spec:
           --output-file "$JSON_RESULTS_FILE" \
           --repo-path "$(workspaces.source.path)" \
           --suites "$(params.test_suites)" $EXTRA_ARGS \
-          "$(params.operator_name)" "$(params.bundle_version)"
+          "$(params.operator_name)" "$(params.bundle_version)" \
+          "$(params.affected_catalog_operators)"
 
         jq . <$JSON_RESULTS_FILE
         jq -r '(.outputs//[])|length' <$JSON_RESULTS_FILE \
--- a/operator-pipeline-images/operatorcert/entrypoints/static_tests.py
+++ b/operator-pipeline-images/operatorcert/entrypoints/static_tests.py
@@ -5,7 +5,7 @@ import json
 import logging
 from typing import Any, Dict, List, Optional
 
-from operator_repo import Repo
+from operator_repo import Repo, OperatorCatalogList
 from operator_repo.checks import Fail, run_suite
 from operatorcert.logger import setup_logger
 from operatorcert.utils import SplitArgs
@@ -49,19 +49,21 @@ def setup_argparser() -> argparse.ArgumentParser:
     parser.add_argument("--verbose", action="store_true", help="Verbose output")
     parser.add_argument("operator")
     parser.add_argument("bundle")
+    parser.add_argument("affected_catalogs")
 
     return parser
 
 
-def check_bundle(
+def execute_checks(  # pylint: disable=too-many-arguments
     repo_path: str,
     operator_name: str,
     bundle_version: str,
+    affected_catalogs: str,
     suite_names: List[str],
     skip_tests: Optional[List[str]] = None,
 ) -> Dict[str, Any]:
     """
-    Run a check suite against the given bundle and return warnings and
+    Run a check suite against the given target and return warnings and
     failures in a format inspired by the operator-sdk bundle validate
     tool's JSON output format
 
@@ -69,6 +71,7 @@ def check_bundle(
         repo_path (str): path to the root of the operator repository
         operator_name (str): name of the operator
         bundle_version (str): version of the bundle to check
+        affected_catalogs (str): Coma separated list of affected catalogs
         suite_name (str): name of the suite to use
         skip_tests (Optional[List]): List of checks to skip
 
@@ -77,13 +80,25 @@ def check_bundle(
     """
     repo = Repo(repo_path)
     operator = repo.operator(operator_name)
-    bundle = operator.bundle(bundle_version)
+    bundle = operator.bundle(bundle_version) if bundle_version else None
+    catalogs = OperatorCatalogList(
+        [
+            catalog
+            for catalog in operator.all_operator_catalogs()
+            if catalog in affected_catalogs.split(",")
+        ]
+    )
 
     outputs = []
     passed = True
 
     for suite_name in suite_names:
-        for result in run_suite([bundle, operator], suite_name, skip_tests=skip_tests):
+        for result in run_suite(
+            # do only catalog checks if bundle is not provided
+            [bundle, operator] if bundle else [catalogs],
+            suite_name,
+            skip_tests=skip_tests,
+        ):
             if isinstance(result, Fail):
                 passed = False
             item = {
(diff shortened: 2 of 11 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/run-static-tests.yml
```diff
@@ -86,7 +89,8 @@ spec:
           --output-file "$JSON_RESULTS_FILE" \
           --repo-path "$(workspaces.source.path)" \
           --suites "$(params.test_suites)" $EXTRA_ARGS \
-          "$(params.operator_name)" "$(params.bundle_version)"
+          "$(params.operator_name)" "$(params.bundle_version)" \
+          "$(params.affected_catalog_operators)"
```
Allda, line 93: Just making sure here.. Did you test if this also works for non-catalog changes? Will an empty string be handled correctly?

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
```diff
@@ -77,13 +80,25 @@ def check_bundle(
     """
     repo = Repo(repo_path)
     operator = repo.operator(operator_name)
-    bundle = operator.bundle(bundle_version)
+    bundle = operator.bundle(bundle_version) if bundle_version else None
+    catalogs = OperatorCatalogList(
+        [
+            catalog
+            for catalog in operator.all_operator_catalogs()
```
Allda, line 87: I am not sure if I understand this loop. The `affected_catalogs.split(",")` returns a list of strings, right? The `operator.all_operator_catalogs()` returns a list of `OperatorCatalog`. And then you compare if `OperatorCatalog` in a list of strings. How it this going to work? Or am I missing anything here?

Also if you already have a list of affected catalogs, why don't you initiate it directly using `OperatorCatalog(<path_to_operator_catalog>)` constructor and instead use an operator and its edge to get to it? If we ever make these types completely independent (operator and operatorCatalog) it won't work.

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
```diff
@@ -77,13 +80,25 @@ def check_bundle(
...
+    )
 
     outputs = []
     passed = True
 
     for suite_name in suite_names:
-        for result in run_suite([bundle, operator], suite_name, skip_tests=skip_tests):
+        for result in run_suite(
+            # do only catalog checks if bundle is not provided
+            [bundle, operator] if bundle else [catalogs],
```
Allda, line 98: Why do we make it mutually exclusive? Can we run all checks if the bundle, operator, and catalog are available?

operator-pipeline-images/operatorcert/static_tests/common/catalog.py
```diff
@@ -0,0 +1,55 @@
...
+    """Filter out invalid images"""
+    invalid_images = []
+    for image in bundle_images:
+        if not any(image.startswith(registry) for registry in allowed_registries):
+            invalid_images.append(image)
+
+    return invalid_images
+
+
+def check_bundle_images_in_fbc(
```
Allda, line 34: Should this file be renamed to `operator_catalogs.py` since it uses `OperatorCatalogList` as an input?

operator-pipeline-images/operatorcert/static_tests/common/catalog.py
```diff
@@ -0,0 +1,55 @@
...
+    operator_catalogs: OperatorCatalogList,
+) -> Iterator[CheckResult]:
+    """Validate if bundle image reference in fbc is in allowed registry"""
+
+    allowed_registries = None
+    for catalog in operator_catalogs:
+        if not isinstance(allowed_registries, list):
```
Allda, line 41: The config is always the same for all catalogs. It doesn't make sense to me to have this in the `for` loop. It unnecessarily pollutes a code and it is hard to read. You can create a separate function to get the list of allowed catalogs and return much earlier if not present. The for loop should really focus on iterating over catalogs and checking values.

operator-pipeline-images/operatorcert/static_tests/common/catalog.py
```diff
@@ -0,0 +1,55 @@
...
+                "allowed_bundle_registries", []
+            )
+        if not allowed_registries:
+            # nothing to check if no registry restriction is defined
+            break
+        bundle_images = _get_bundle_registries_from_catalog(catalog)
+        invalid_images = _filter_invalid_images(bundle_images, allowed_registries)
+        if invalid_images:
+            yield Fail(
+                f"Invalid bundle image(s) found in {str(catalog)}: "
```
Allda, line 53: Let's make this error more user-friendly. Give a user at least information on which registries are allowed. If I were a user reading this message and not knowing about the purpose of this test I wouldn't know why it failed.

ansible/roles/operator-pipeline/templates/openshift/tasks/run-static-tests.yml
mantomas: I tested non-catalog changes by running integration tests manually [here](https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/2040). The test already work with empty string when `bundle_version` is empty. It also work for catalog changes, so I expect empty strings are handled correctly and passed as positional arguments.

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
mantomas: Operator catalog is [comparable to string](https://github.com/mporrato/operator-repo/blob/277471a08a7e9f06169925fcf25b039ac70c4677/src/operator_repo/core.py#L599) and it works both ways. 

I don't understand what you mean by `If we ever make these types completely independent`? Is there such a plan? As of now, we always have an Operator (as it is affected by the catalog change iirc) and using the edge give me all its catalogs. I can see that there is the way to get specific catalog like `OperatorCatalog(repo.root / "catalogs" / "v4.13/kiali")` - for me having the path already provided by the Operator sounds more straightforward, but I am open to discussion.

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
mantomas: As of now we have validation preventing simultaneous bundle and catalog changes, so we don't have both at the same time. And for current workflow, operator check was also skipped if bundle was not affected.
If it makes sense, to validate at least the operator together with the catalog, I can change the line to something like `[bundle, operator] if bundle else [catalogs, operator]` to have operator covered in both workflows.

operator-pipeline-images/operatorcert/static_tests/common/catalog.py
mantomas: This structure is kinda hard-coded to the [operator-repo](https://github.com/mporrato/operator-repo/blob/277471a08a7e9f06169925fcf25b039ac70c4677/src/operator_repo/checks/__init__.py#L84) library. I can change it there though so we can decide in the future if there will be any validation for the whole catalog.

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
mantomas: Changed the logic to use `repo.catalog_path()`

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
mantomas: Changed so Operator checks are run in both cases.
As it is now, the `run_suite()` expects specific class instances and providing `None` instead of `Bundle` leads to an exception.

operator-pipeline-images/operatorcert/static_tests/common/catalog.py
mantomas: Fixed in external `operator-repo` and renamed.

operator-pipeline-images/operatorcert/static_tests/common/catalog.py
mantomas: Good point - moved outside of the `for` loop.

mantomas:
I did some updates and to verify the results run the pipeline for [FBC](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/317) and also [integration tests](https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/2057) - both with expected results.

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
mantomas: Last update changes the logic to run each suite available.

operator-pipeline-images/operatorcert/entrypoints/static_tests.py
mantomas: @Allda I reverted this logic back to use the `operator` to `operator_catalogs` edge - the `affected_catalogs` contain also removed catalogs and this way we can filter them out. Initiate catalog directly `OperatorCatalog(<path_to_operator_catalog>)` raise exception for deleted operator catalogs.

Allda (approved):

tomasbakk (approved):

PR #758, opened 2024-12-11
Skip yamllint for removed bundles

Allda:
In case a bundle is removed the files don't exist in the repository
anymore and it doesn't make sense to run yamllint.

The change parser also needs to distinguish between added, modified, and
removed operators and bundles. Each of these categories needs to be
treated differently.

JIRA: ISV-5466

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/read-config.yaml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/read-config.yaml
@@ -41,7 +41,13 @@ spec:
 
         CONFIG_PATH="$PKG_PATH/ci.yaml"
 
-        if [[ ! -f "$CONFIG_PATH" ]]; then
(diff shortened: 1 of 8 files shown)


mantomas (approved):
lgtm

ansible/roles/operator-pipeline/templates/openshift/tasks/read-config.yaml
```diff
@@ -41,7 +41,13 @@ spec:
 
         CONFIG_PATH="$PKG_PATH/ci.yaml"
 
-        if [[ ! -f "$CONFIG_PATH" ]]; then
+        if [[ ! -d "$PKG_PATH" ]]; then
+            echo "Operator directory is removed."
```
ezopezo, line 45: Non-blocking: Does it make a sense to indicate here which operator has been removed?

ezopezo (approved):

PR #765, opened 2025-01-07
Add helper functions for bundle image

Allda:
A new class BundleImage is introduced that represents a bundle image. The class will be used in integration test to verify its content is as expected. The class contains a generic method to get a metadata and simplifies a way how to extract it.

JIRA: ISV-5280

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/operator-pipeline-images/operatorcert/bundle.py
+++ b/operator-pipeline-images/operatorcert/bundle.py
@@ -1,0 +1,215 @@
+"""A module for working with bundle images."""
+
+import json
+import logging
+import os
+import shutil
+import tempfile
+from typing import Any, Optional
+
+import yaml
+from operatorcert import utils
+
+LOGGER = logging.getLogger("operator-cert")
+
+
+class BundleImage:
+    """
+    A class representing a bundle image. The class provides methods to work with
+    the bundle image content.
+    """
+
+    def __init__(
+        self, image_pullspec: str, auth_file_path: Optional[str] = None
+    ) -> None:
+        self.image_pullspec = image_pullspec
+        self.auth_file_path = auth_file_path
+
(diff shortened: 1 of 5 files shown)


operator-pipeline-images/operatorcert/bundle.py
```diff
@@ -0,0 +1,215 @@
...
+    def __repr__(self) -> str:
+        """
+        A representation of a BundleImage object.
+
+        Returns:
+            str: A representation of a BundleImage object.
+        """
+        return f"BundleImage({self.image_pullspec})"
+
+    def __del__(self) -> None:
```
ezopezo, line 49: Have you considered that destructors are very unsafe to use, because there is no guarantee when garbage collection invoke them (and thus in this implementation when will be the file removed)?
If you need to be sure about when file will be removed and avoid potential future inconsistencies or race conditions I strongly recommend to use rather context manager pattern (\_\_enter\_\_ and \_\_exit\_\_ methods) and use BundleImage as context manager.

Allda: I don't care when exactly this file gets removed; I'll leave it up to Python. The important thing is that the file is available when the object is usable. Since it uses random temporary directories, it never conflicts with any other instance of the same class. I didn't want to overcomplicate the testing scenarios with a context manager if there was no specific need for it.

ezopezo (approved):

operator-pipeline-images/operatorcert/bundle.py
ezopezo: Fair enough.

mavaras (approved):

PR #772, opened 2025-02-05
[ISV-5276] Create helper functions for GitHub branches for copy and delete operations.

haripate:
### Merge Request Checklists

    - [x] Linting
    - [x] Code formatter - Black
    - [x] Security scanners
    - [x] Unit tests
    - [ ] Integration tests
- [x] Code is reviewed by at least 1 team member


operator-pipeline-images/operatorcert/github.py
```diff
@@ -336,3 +336,81 @@ def close_pull_request(
...
+    """
+    src_repository = github_client.get_repo(src_repo_name)
+    dest_repository = github_client.get_repo(dest_repo_name)
+
+    src_branch_ref = src_repository.get_branch(src_branch_name)
+    latest_commit_sha = src_branch_ref.commit.sha
+
+    dest_branch_exists = any(
+        branch.name == dest_branch_name for branch in dest_repository.get_branches()
+    )
```
mantomas, line 366: Isn't there any option to do this check without iterating all branches? Maybe getting the dest branch the same way as the src and act upon what it returns? Not sure how the library works in case object does not exist - all the previous calls to get_repo/branch might also need some kind of try/except mechanism to fail correctly if objects don't exist I think.

operator-pipeline-images/operatorcert/github.py
```diff
@@ -336,3 +336,81 @@ def close_pull_request(
...
+        github_client (Github): A Github API client
+        repository_name (str): A repository name in the format "organization/repository"
+        branch_name (str): The name of the branch to delete
+    """
+    repository = github_client.get_repo(repository_name)
+    branch_ref = f"heads/{branch_name}"
+
+    branch_exists = any(
+        branch.name == branch_name for branch in repository.get_branches()
+    )
```
mantomas, line 408: Same as the other comment - what if the repo has hundreds of test branches? I tested the the `get_branch()' with nonexistent branch and it raises GithubException, so maybe try/except might work better here?

operator-pipeline-images/operatorcert/github.py
mporrato: I agree with Tomas' comment about error handling but, since the rest of the module does not have any retry mechanism, I think this can be done as a future improvement.

Also, I don't think we need to handle the case where the destination branch already exists: I would be perfectly ok raising an exception instead of updating the existing branch.

mporrato (changes requested):
Have you tried to copy a branch between two different repos with different histories (as opposed to making a duplicate of an existing branch on a specific repo)? Does it work?

operator-pipeline-images/operatorcert/github.py
haripate: Understood. I initially used a try-except block, but switched to a different approach as the Python mypy unit tests were failing due to common exception errors. I will now revert to try-except.

harita96:
> Have you tried to copy a branch between two different repos with different histories (as opposed to making a duplicate of an existing branch on a specific repo)? Does it work?

@mporrato I updated the PR with new changes where the copy operation between two repo with different history is possible. I tested it in my local system and it is working.

operator-pipeline-images/operatorcert/github.py
```diff
@@ -336,3 +343,117 @@ def close_pull_request(
...
+            if isinstance(contents, list):
+                for content in contents:
+                    collect_files_recursive(content.path)
+            else:
+                file_data = contents.decoded_content
+                element = InputGitTreeElement(
+                    path=contents.path,
+                    mode="100644",
+                    type="blob",
+                    content=file_data.decode("utf-8"),
```
mporrato, line 395: This won't work with files that are not regular files (for example links) or generic binary files (which cannot be decoded as utf-8 text).
This is not ideal and can cause issues, but I don't think we can fix this without cloning the source repo locally an pushing to the destination; however this might be enough for what we need to achieve in the short term and we may want to fix this down the road.
Can you please note this in the docstring?

haripate: Updated the doc string with the limitation details.

mporrato (approved):

mantomas (approved):

PR #775, opened 2025-02-06
FBC make uses catalog mapping instead of hardcoded version

Allda:
A makefile that generates FBC catalogs was previously using hard coded ocp version. This mechanism has been replaced by the new template to catalog mapping available in ci.yaml file.

When rendering a catalog using Makefile the mapping determines which catalogs are produced.

JIRA: ISV-5505

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/fbc/Makefile
+++ b/fbc/Makefile
@@ -4,12 +4,6 @@
 # The makefile should be placed in the root of the operator repository.
 # for example at: <operator-repo>/operators/<operator-name>/Makefile
 
-# A user can customize "catalog" target to generate the operator catalog in a way
-# that suits the operator.
-# OPM allows for the generation of catalogs using different templates.
-# - basic: generates a basic catalog
-# - semver: generates a catalog with semver versioning
-
 PWD=$(shell pwd)
 OPERATOR_NAME=$(shell basename $(PWD))
 TOPDIR=$(abspath $(dir $(PWD))/../)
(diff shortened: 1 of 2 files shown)


fbc/Makefile
```diff
@@ -61,60 +46,56 @@ fbc-onboarding: clean
...
 
 # validate-catalogs target illustrates FBC validation
 # all FBC must pass opm validation in order to be able to be used in a catalog
 .PHONY: validate-catalogs
 validate-catalogs: ${BINDIR}/opm
-	for version in $(OCP_VERSIONS); do \
-		${BINDIR}/opm validate $(CATALOG_DIR)/$${version}/${OPERATOR_NAME} && echo "$${version} catalog validation passed" || echo "$${version} catalog validation failed"; \
-	done
+	@find ${TOPDIR}/catalogs -type d -name fbc-test-operator -exec \
+		sh -c '${BINDIR}/opm validate $$(dirname "{}") && echo "✅ Catalog validation passed: {}" || echo "❌ Catalog validation failed: {}"' \;
```
acornett21, line 71: This no longer works, it's basically doing a no-op output.

Allda: Why is that? When I make a mistake in the catalog on purpose this gives me the error and the command fails.

RichardPlesnik (approved):

fbc/Makefile
Allda: I had to merge this to not break users with incompatible Makefiles. Let's still discuss this and if we find it is not correct I am going to fix it in next PR.

PR #789, opened 2025-02-19
ISV-5645: Signing pipeline for image bootstrapping

Allda:
A new pipeline is created that will server for the automating of index image bootstrapping process. The new pipeline allow signing only images that haven't been configured in Pyxis as GA.

The new pipeline will be deployed into separate namespace which will be restricted only to a service account that can trigger a pipeline.

JIRA: ISV-6546

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests


ansible/roles/operator-pipeline/templates/openshift/tasks/index-signing-image-check.yml
```diff
@@ -0,0 +1,88 @@
...
+        PYXIS_BASE_URL="https://catalog.redhat.com/api/containers"
+        if [ "$PYXIS_ENV" != "prod" ]; then
+          PYXIS_BASE_URL="https://catalog.stage.redhat.com/api/containers"
+        fi
+
+        FULL_PYXIS_URL="$PYXIS_BASE_URL/v1/operators/indices?filter=path==$(params.image_pullspec)"
+
+        # Use squid proxy to access non prod environments
+        curl -x http://squid.corp.redhat.com:3128 -s -X GET $FULL_PYXIS_URL > /tmp/pyxis_response.json
+        cat /tmp/pyxis_response.json | jq .
```
mporrato, line 48:
```suggestion
        jq . /tmp/pyxis_response.json
```

ansible/roles/operator-pipeline/templates/openshift/tasks/index-signing-image-check.yml
```diff
@@ -0,0 +1,88 @@
...
+
+        export ITEMS=`cat /tmp/pyxis_response.json | jq -r ".data | length"`
```
mporrato, line 50:
```suggestion
        export ITEMS=$(jq -r ".data | length" /tmp/pyxis_response.json)
```

ansible/roles/operator-pipeline/templates/openshift/tasks/index-signing-image-check.yml
```diff
@@ -0,0 +1,88 @@
...
+        PYXIS_ENV=$(params.pyxis_env)
+        PYXIS_BASE_URL="https://catalog.redhat.com/api/containers"
+        if [ "$PYXIS_ENV" != "prod" ]; then
+          PYXIS_BASE_URL="https://catalog.stage.redhat.com/api/containers"
+        fi
+
+        FULL_PYXIS_URL="$PYXIS_BASE_URL/v1/operators/indices?filter=path==$(params.image_pullspec)"
+
+        # Use squid proxy to access non prod environments
+        curl -x http://squid.corp.redhat.com:3128 -s -X GET $FULL_PYXIS_URL > /tmp/pyxis_response.json
```
mporrato, line 47: Even before calling the pyxis api, shouldn't we check that the pullspec (excluding the tag) points to a valid index? Otherwise the pyxis call would return nothing and therefore the pipeline would go ahead with the signing.

ansible/roles/operator-pipeline/tasks/index-img-bootstrap-signing.yml
```diff
@@ -0,0 +1,133 @@
...
+- name: Deploy pipeline tasks
+  tags:
+    - tekton-task
+  kubernetes.core.k8s:
+    state: present
+    apply: true
+    namespace: "{{ oc_index_bootstrap_namespace }}"
+    definition: "{{ lookup('template', '{{ item }}') }}"
+  with_items:
+    - ../templates/openshift/tasks/set-env.yml
```
mporrato, line 120: Looks like the set-env.yml file is missing from the PR.

mporrato (commented):
The two vault-encrypted files are different but contain the exact same key material and they have credentials for for both registry.redhat.io and registry.stage.redhat.io: it does not look right.

ansible/roles/operator-pipeline/templates/openshift/tasks/index-signing-image-check.yml
Allda: Fixed

ansible/roles/operator-pipeline/templates/openshift/tasks/index-signing-image-check.yml
Allda: Fixed

ansible/roles/operator-pipeline/templates/openshift/tasks/index-signing-image-check.yml
Allda: That's one thing I thought I added but forgot. The task now contains a check for allowed repositories.

jedinym (approved):

Allda:
> The two vault-encrypted files are different but contain the exact same key material and they have credentials for for both registry.redhat.io and registry.stage.redhat.io: it does not look right.

I originally thought we would be supporting a use case for signing non-prod images in a prod environment but we can drop this off. I removed the nonprod token from prod secret. However I kept both in nonprod secret for testing purposes. The reason why they are different are because we have different vault passphrases for prod and nonprod.

docs/bootstrap-signing.md
```diff
@@ -0,0 +1,66 @@
+# Index bootstrap signing
+
+Whenever there is a new release of the OpenShift we have to release a new version of operator catalog.
+This process involves several steps that involves several people and teams.
+To reduce amount of manuall work we automated the process. One of the steps in the
```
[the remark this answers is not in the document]
Allda: Fixed

PR #794, opened 2025-02-26
fix: merge default and channels in check_pruned_graph check

Allda:
The check skipped the default channel when checking for pruned update graph. This caused the test to skip certain corner cases when pipeline let graph to be pruned.

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/operator-pipeline-images/operatorcert/static_tests/isv/bundle.py
+++ b/operator-pipeline-images/operatorcert/static_tests/isv/bundle.py
@@ -32,6 +32,9 @@ def check_pruned_graph(bundle: Bundle) -> Iterator[CheckResult]:
 
     if skip_range and not replaces:
         channels = bundle.channels
+        if bundle.default_channel:
+            channels.add(bundle.default_channel)
+
         operator = bundle.operator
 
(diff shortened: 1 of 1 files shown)


acornett21 (approved):

Allda:
@mporrato I know you had some comments on a Slack about this PR. Can you please agree with @acornett21 on whether this is good to be merged or if we need to close it? Thanks

mporrato:
@Allda my concern in the slack thread was about the [docs](https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md) which, in my opinion, suggests that the default channel annotation does not apply to the individual bundle but to the entire operator: if that's the case, this change would at least cause some false negatives. I suggested that the root cause of the issue under discussion might have been the order in which the PRs were merged.
If you guys do not agree with my interpretation, by all means, please go ahead and merge this.

Allda:
ok, closing it for now.

PR #796, opened 2025-03-04
Refactor signing tasks

Allda:
The reason for this refactoring is that tekton results and params have a size limit that can be reached in case we enable another OCP version. The refactoring replaces results with workspaces. The workspace is used to transfer data between tasks.

The summary of this change is that we merged signing related tasks into single tasks. This simplifies the pipeline and make it faster since it deploy less pods.

The refactoring also enables removing the "decide-index-paths" because the logic is no longer needed.

JIRA: ISV-5683

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests


mantomas (approved):
LGTM

ezopezo (approved):

PR #816, opened 2025-05-09
Add option to specify basic-auth workspace for checkout task in operator ci pipeline

JoelRuizRojas:
Even when it is recommended the git authentication using ssh, basic authentication should be fully supported. It was missing to propagate the `basic-auth` workspace from the pipeline to the task.

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
@@ -73,6 +73,8 @@ spec:
     - name: pipeline
     - name: ssh-dir
       optional: true
+    - name: basic-auth
+      optional: true
     - name: registry-credentials
       optional: true
     - name: registry-cacert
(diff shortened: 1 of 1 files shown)


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
```diff
@@ -73,6 +73,8 @@ spec:
     - name: pipeline
     - name: ssh-dir
       optional: true
+    - name: basic-auth
```
Allda, line 76: Please rename this workspace to something meaningful in the context of the pipeline. For example, `git_repo_basic_auth` or similar. Please also add a description for this workspace so it is obvious what it is used for. Thanks.

JoelRuizRojas: @Allda renamed the workspace as suggested but only at the Pipeline level which impacts the user 6c671031ba2039a006c850b9cd7a534a40e9f22d. The inner name still the same. If you think we should do a rename everywhere please let me know. I tested it like this and it works.

PR #819, opened 2025-05-15
FBC auto-release uses migration option for 4.17+ catalogs

Allda:
The command used for opm template rendering requires a different set of options for catalogs 4.17+. For these catalogs we need to use "--migrate-level bundle-object-to-csv-metadata" in order to produce valid catalogs.

JIRA: ISV-5919

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/operator-pipeline-images/operatorcert/entrypoints/add_bundle_to_fbc.py
+++ b/operator-pipeline-images/operatorcert/entrypoints/add_bundle_to_fbc.py
@@ -5,9 +5,10 @@ A module responsible for automated release of bundle to the FBC catalog.
 import argparse
 import logging
 import os
+import subprocess
 from abc import ABC, abstractmethod
-from typing import Any
 from pathlib import Path
+from typing import Any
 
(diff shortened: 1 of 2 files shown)


wcheang:
The code looks fine to me, but since I'm not familiar with the FBC workflow, just wanted to make sure: is it safe to assume that using a cache that never expires is fine, since it's only ever saved in the taskrun?

Allda:
> The code looks fine to me, but since I'm not familiar with the FBC workflow, just wanted to make sure: is it safe to assume that using a cache that never expires is fine, since it's only ever saved in the taskrun?

Yes that's fine. The cache is only used to reduce the number of calls if we render a catalog from the same template. It will always run in a context of the same operator bundle and after a task is complete a cache is wiped.

wcheang (approved):

tomasbakk (approved):

PR #838, opened 2025-07-07
[ISV-5281] [ISV-5282] Integration tests

mporrato:
- Implemented integration test main loop
- Partially implemented a basic test case

### Merge Request Checklists

    - [x] Linting
    - [x] Code formatter - Black
    - [x] Security scanners
    - [x] Unit tests
    - [x] Integration tests

### DIFF
--- a/integration-tests-config-sample.yaml
+++ b/integration-tests-config-sample.yaml
@@ -1,5 +1,9 @@
 ---
 
+fixtures_repository:
+    # The GitHub repo containing fixtures used by test cases
+    url: https://github.com/foo/operators-integration-tests-fixtures
+    token: secretABC
 operator_repository:
     # The GitHub repository hosting the operators for integration tests
     url: https://github.com/foo/operators-integration-tests
--- a/operator-pipeline-images/operatorcert/integration/testcases/basic.py
+++ b/operator-pipeline-images/operatorcert/integration/testcases/basic.py
@@ -1,0 +1,82 @@
+"""
+Basic test cases
+"""
+
+import logging
+from pathlib import Path
+from shutil import rmtree
+from tempfile import mkdtemp
+from typing import Optional
+
+from git import Repo
+
+from operatorcert.integration.config import Config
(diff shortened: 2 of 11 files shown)


integration-tests-config-sample.yaml
```diff
@@ -1,5 +1,9 @@
 ---
 
+fixtures_repository:
+    # The GitHub repo containing fixtures used by test cases
+    url: https://github.com/foo/operators-integration-tests-fixtures
```
Allda, line 5: Does this already exist, at least for your internal testing?

operator-pipeline-images/operatorcert/integration/testcases/basic.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+        # configure remotes
+        local_repo.create_remote("operators", url=self.config.operator_repository.url)
+        local_repo.create_remote(
+            "contributor", url=self.config.contributor_repository.url
+        )
+        # The fixture branch contains the operator structure needed for the test
+        # and the latest commit is the change being tested. This means we need
+        # to exclude the latest commit when pushing to the operator repo.
+        local_repo.create_head(
```
Allda, line 44: This is not something that should be fixed as part of this PR, but we should be able to dynamically change a version of an operator instead of using a hardcoded version for all test-cases.

RichardPlesnik (approved):

integration-tests-config-sample.yaml
mporrato: No, it's supposed to contain ideally one branch per test case where each branch contains the minimum necessary to perform the test: in this case it would be an existing operator with at least one bundle and a commit (the head of the branch) would be the change to be tested.

operator-pipeline-images/operatorcert/integration/testcases/basic.py
mporrato: This creates the branch in the local (ephemeral) copy of the repo just for convenience in the push step below; it's isolated from other tests so there is no chance of collision.

PR #839, opened 2025-07-28
Add code for webhook dispatcher

Allda:
A dispatcher is a new component to manage incoming webhooks from Github. It accept new requests and manage a queue based on cluster capacity.

A capacity is defined as a number of running pipelines in certain namespace. This simple metrics will provide a buffer mechanism for time periods when we receive a bunk releases of operators. The mechanism keeps pending requests in the database queue and only trigger related pipeline in case of free capacity.

The solution is made of:
- Rest API
- Postgres database
- Dispatcher
- Capacity manager

JIRA: ISV-6108

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests


operator-pipeline-images/operatorcert/webhook_dispatcher/dispatcher.py
```diff
@@ -0,0 +1,237 @@
...
+            event.repository_full_name,
+        )
+        if not await pipeline_event.is_capacity_available():
+            LOGGER.debug(
+                "There is not enough capacity to process the event %s", event.id
+            )
+            # Waiting for a next loop to process the event
+            return
+
+        await pipeline_event.trigger_pipeline()
```
RichardPlesnik, line 169: Shouldn't some error handling be added here so the event is not marked as processed even if the HTTP call doesn't return 202 and `trigger_pipeline` function returns False?

RichardPlesnik (approved):
One more nitpick, otherwise LGTM :+1: Good job with the design!

PR #843, opened 2025-08-07
Add retry mechanism for registry push commands

Allda:
There are 2 types of commands that are handled by this commit.

1. Bash commands executed in the Tekton task
2. Commands executed in the Python script

Both of these uses re-tries mechanism now. The bash commands use a generic script that controls the execution and retries if needed.

This commit "fixes" the tasks and commands that failed the most often and improves stability.

JIRA: ISV-6114

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/operator-pipeline-images/operatorcert/utils.py
+++ b/operator-pipeline-images/operatorcert/utils.py
@@ -134,33 +134,43 @@ class SplitArgs(argparse.Action):
 
 
 def run_command(
-    cmd: List[str], check: bool = True, cwd: Optional[str] = None
+    cmd: List[str], check: bool = True, cwd: Optional[str] = None, retries: int = 1
 ) -> subprocess.CompletedProcess[bytes]:
     """
     Run a shell command and return its output.
 
     Args:
         cmd (List[str]): Command to run
+        check (bool): Whether to check the command exit code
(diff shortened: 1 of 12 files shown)


operator-pipeline-images/operatorcert/utils.py
```diff
@@ -134,33 +134,42 @@ def __call__(
...
-        raise e
+    for attempt in range(1, retries + 1):
+        try:
+            output = subprocess.run(
+                cmd,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                check=check,
+                cwd=cwd,
+            )
```
mantomas, line 160: Shouldn't there be a `break` statement at the end of the try block? Or what happens if there is successful attempt before the retries are exhausted?

mantomas (commented):
Overall it looks good to me, with one qustion.

operator-pipeline-images/operatorcert/utils.py
Allda: Good catch! It is fixed and covered by the unit tests now.

PR #858, opened 2025-09-25
[ISV-6166] add CSV parsing before add-image-to-bundle step

tomasbakk:
I reorganised the tasks in the hosted pipeline in order for preflight to validate and catch the incorrectly referenced CSV in the earlier stages. There should not be any problems by doing so. The hosted pipeline can run it's course normally.

Preflight is using 4 validators from operator-sdk -> https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/internal/bundle/bundle.go#L50-L55 -> 2 of them are responsible for metadata requirements, 1 is checking API deprecation and `OperatorHubV2Validator` is validating the bundle against the operatorhub.io requirements which is one we need. (Validators (all and ones used in preflight) are referenced here in operator-framewrok api: https://github.com/operator-framework/api/blob/master/pkg/validation/validation.go)

If there are incorrectly referenced specs in CSV (such as was the case why this ticket was raised), not only will the validation fail for checking the operatorhub requirements but other scorecards will fail as well.

Since the bundle will be run through validations before add-bundle-to-index task, we will save up the time and owners can fix potential problems earlier, with a clear error message.

Tested in preprod; artifacts from preflight -> [logs](https://gist.github.com/rh-operator-bundle-bot/09a7b019ae81d1297fddb1843cfc8206) and pipeline [logs](https://gist.github.com/rh-operator-bundle-bot/bca44dae35596e55e70f3a64898ebb93).

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-hosted-pipeline.yml
@@ -821,139 +821,11 @@ spec:
           workspace: repository
           subPath: src
 
-    - name: build-fbc-index-images
-      runAfter:
-        - get-supported-versions
-      when:
-        - *catalogFileChange
-      taskRef:
-        name: build-fbc-index-images
-      params:
-        - name: pipeline_image
-          value: "$(params.pipeline_image)"
-        - name: index_images
-          value: "$(tasks.get-supported-versions.results.indices)"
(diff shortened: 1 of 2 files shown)


tomasbakk:
Closing PR as this change will not be implemented. Details in the jira story itself.

PR #907, opened 2026-03-12
fix: require approval for all steps

bclindner:
potential for PwnRequest attacks is low but not zero - this should
ensure things are safe going forward

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -14,6 +14,7 @@ jobs:
   tox:
     name: Run unit tests and linters
     runs-on: ubuntu-latest
+    environment: ${{ github.event.pull_request.head.repo.fork && 'build-and-test' || '' }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2


JakubDurkac:
Good job with spotting the weakness. As I understand it, now for each MR that comes from a fork there's an approval needed by people defined in the build-and-test environment, that for each tox job run. I agree this solves the security issue and the change makes sense to me and I'm for it. Just one thing I wonder about a little bit is if this could become a bit annoying if someone wants to create something new and would be failing tox tests in pipeline a few times and needed approval for each time from certain people. But I guess if developers make sure to successfully run tox locally before contributing, then it shouldn't be an issue. Maybe let's wait also for what Richard has to say about this.

RichardPlesnik:
Good catch, Brian! LGTM :+1: 

I agree with Jakub, that it could be limiting for the some contributors, but they are supposed to run tox test locally anyway. It is definitely worth the discomfort.

RichardPlesnik (approved):

RichardPlesnik:
I'm also wondering, do we need to rotate some secrets just to be sure we're safe?

Can you check whether any PR tried to use this vulnerability, if you haven't done so already? 

If it is more complex, feel free to create a separate ticket for the investigation.

JakubDurkac (approved):

bclindner:
> I'm also wondering, do we need to rotate some secrets just to be sure we're safe?

That might be worth ticketing out just to be safe; might be easy but I'm saying ticket b/c I'm not sure what the process is lol

> Can you check whether any PR tried to use this vulnerability, if you haven't done so already?

There's no way to know for sure w/o going through every PR we've ever done, since this has been a known attack vector for years, but I can say the last several months of changes have come from internal RH contributors, so I think we're OK there.

PR #921, opened 2026-03-30
feat(ISV-6696): Hosted pipeline uses pending index repositories

JakubDurkac:
Switch to using pending repositories instead of public repository mirrors as source for index images. Merging to non-primary branch, so that we can test and merge both this (ISV-6696) and release pipeline changes (ISV-6697) to main at the same time.

### Merge Request Checklists

- [ ] Development is done in feature branches
- [ ] Code changes are submitted as pull request into a primary branch [Provide reason for non-primary branch submissions]
- [ ] Code changes are covered with unit and integration tests.
- [ ] Code passes all automated code tests:
    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests
- [ ] Code is reviewed by at least 1 team member
- [ ] Pull request is tagged with "risk/good-to-go" label for minor changes

PR #923, opened 2026-04-01
feat(ISV-6697): Release pipeline uses pending index repositories

JakubDurkac:
1. For release pipeline, use pending repositories as source for index images.
2. Use IIB overwrite token in release pipeline to overwrite the source pending repositories.
3. Specify tags (version + permanent tag) for index images. These are passed to IIB API calls and publish task.
4. Passing the overwrite token using newly added "iib-credentials" workspace. Added 'set -e' and switched to not using '--verbose' for add-bundle-to-index and build-fbc-index-image, the tasks token is newly passed to (inspired by publish-task and add-bundle-to-index, which worked with secrets prior to these changes).

Merging to non-primary branch, so that we can test and merge both this (ISV-6696) and release pipeline changes (ISV-6697) to main at the same time.

WARNING: Currently it seems that the overwrite token passed to IIB API build request is not working due to some bug, therefore the builds fail, because they fail to overwrite "-pending" repositories (unauthorized), so we can't fully test the changes currently. More info here in the [ticket](https://redhat.atlassian.net/browse/CLOUDDST-32419) I've opened.

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests


ansible/roles/operator-pipeline/templates/openshift/tasks/add-bundle-to-index.yml
```diff
@@ -97,6 +97,17 @@ spec:
           EXTRA_ARGS+=" --mode $(params.upgrade-graph-mode)"
         fi
 
+        # Add IIB overwrite token if credentials workspace is provided (for release pipeline)
+        if [[ "$(workspaces.iib-credentials.bound)" == "true" ]]; then
+          IIB_QUAY_USER=$(cat $(workspaces.iib-credentials.path)/username)
+          IIB_QUAY_TOKEN=$(cat $(workspaces.iib-credentials.path)/password)
+          EXTRA_ARGS+=" --iib-overwrite-token ${IIB_QUAY_USER}:${IIB_QUAY_TOKEN}"
```
Allda: We should avoid using cli params to pass secrets. I think better solution is to use env vars and load it from the env vars in the script.

JakubDurkac: Fixed in the latest commit. Let me know if you meant it differently.

operatorcert/entrypoints/index.py
```diff
@@ -2,10 +2,12 @@
 IIB module for building a index images for a bundle
 """
 
+# pylint: disable=duplicate-code
```
RichardPlesnik: Can this be set only on affected parts of code rather than on the whole file?

JakubDurkac: Should be better now.

ansible/roles/operator-pipeline/templates/openshift/tasks/add-bundle-to-index.yml
```diff
@@ -67,22 +78,11 @@ spec:
           mountPath: "/etc/kerberos"
       script: |
         #! /usr/bin/env bash
-        set -xe
+        # DO NOT USE set -x to avoid revealing credentials in logs!
+        set -e
 
         ENV=$(params.environment)
```
RichardPlesnik: This can be removed, since it is unused after removal of the if condition.

RichardPlesnik: I am not seeing the removal of the condition in diff anymore. Was it something form different PR that is already present in the target branch?

JakubDurkac: Yes, we merged #921 , so ISV-6696 (hosted pipeline changes) to the target branch ISV-6696-6697, and it included removal of the "if" conditions. So the [commit ](https://github.com/redhat-openshift-ecosystem/operator-pipelines/commit/393f70aedc8b749a7f588801d4cc7b178e338713) is already part of target branch.

RichardPlesnik: I see, thanks for explaining :+1:

RichardPlesnik (approved):

JakubDurkac:
As it seems it will take some time for the overwrite token bug to be fixed by IIB, I've made a simple temporary workaround that we can remove later, doing the overwriting manually. Made it just to see if the tests results look otherwise correct. Here are some tests with logs, you can look if you're interested. I was mainly checking if the sources and destinations are as expected for both add-bundle task, and then the publish task, including the version + permanent tags. Those look correct for these tests:

1. make build-and-test-isv
- [MR](https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/3204)
- [Release pipeline logs](https://gist.github.com/rh-operator-bundle-bot/b736701103137eaf2c1d410f0c042d3b)

2. make build-and-test-community
- [MR](https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/3207)
- [Release pipeline logs](https://gist.github.com/rh-operator-bundle-bot/83b01e4bf2b11079b2a0e096b3b71cc7)

3. make build-and-test-isv-fbc-bundle (also successful, but probably not that relevant to the changes made)
- [MR](https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/3209)

The 'make build-and-test-isv-fbc-catalog' is currently getting stuck for me on hosted pipeline, I'm trying to find out why. I'll add results if I get some (EDIT: fixed, see comment below). Just FYI @Allda, @RichardPlesnik. Also,  please let me know if you would like to run some additional tests, so when the IIB is fixed, we are prepared to move forward.

JakubDurkac:
I've found out that the pipeline was getting stuck on fbc catalog test, because get-supported-versions task was generating results that were too big. Apparently by my addition of max_supported_pending_repository we barely hit over 4 KB limit by a few bytes, which lead to the results being truncated. The "message" attribute of "supported-version-check" step, which is a string holding the JSON result object was missing few versions of the last attribute "indices_ocp_versions" and proper closing brackets, which lead to pipeline getting stuck. We would've probably hit it also if there was like 1 more supported OCP version in the list, so to be completely safe long term, I've removed currently unused repositories_with_version and public_repository_mirrors_with_version from results, which should make the results like 60 % smaller. The last thing we need to pass all tests is for IIB to put a new functioning default reading token to stage, because there's a wrong one active at the moment, therefore pipeline is failing with unauthorized, could not find robot with username... logs: https://gist.github.com/rh-operator-bundle-bot/bcae4a6a66a0ac9a0126374839965694

ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -565,7 +584,7 @@ spec:
         - name: pipeline_image
           value: "$(params.pipeline_image)"
         - name: index_images
-          value: "$(tasks.get-supported-versions.results.public_repository_mirrors_with_version)"
+          value: "$(tasks.get-supported-versions.results.max_supported_pending_repository)"
```
Allda: Shouldn't this be pending_repositories_with_version?

Allda (changes requested):
I noticed in the pipeline logs that by changing the IIB build source you accidentally change the signing step. The old solution send a signing request with following payload:
```
"docker_reference": "registry.redhat.io/redhat/community-operator-index:v4.14", "image_name": "redhat/community-operator-index",
```
The new solution is using a different reference:
```
'docker_reference': 'quay.io/redhat-pending/redhat----community-operator-index-pending:v4.20', 'errors': [], 'image_name': 'redhat-pending/redhat----community-operator-index-pending',
```

This needs to be fixed; otherwise, the indexes won't be properly signed.

JakubDurkac:
I made changes to sign-index-image task:
1. In release pipeline, task uses public repository mirror as reference (added public_repository_mirror as parameter passed to the task).
2. In release pipeline, task uses pending repository with digest SHA as inspect target for skopeo (added secret for reading '*-pending' repositories; prod env can read (inspect) quay 'redhat' org's pending repositories; non-prod envs can read quay 'redhat-pending' org's pending repositories).
3. Task is made backward compatible with other pipeline that uses it (index-img-bootstrap-signing-pipeline). Based on presence of public_repository_mirror param, decides to set reference and inspect target for skopeo, either as described above for release pipeline usage, or to reference = inspect target = public repository mirror for index image bootstrap signing pipeline, the same way it was before.

When the certificate renewal is merged to main #933, I'll rebase and run the tests.

JakubDurkac:
The signing step should be fixed now. Let me know what you think. @Allda

Allda (approved):

PR #932, opened 2026-04-16
[ISV-7037] Add defaultChannel support to release-config.yaml schema

haripate:
JIRA Ticket: https://redhat.atlassian.net/browse/ISV-7037

### Merge Request Checklists

    - [x] Linting
    - [x] Code formatter - Black
    - [x] Security scanners
    - [x] Unit tests
    - [x] Integration tests

### DIFF
--- a/operatorcert/entrypoints/add_bundle_to_fbc.py
+++ b/operatorcert/entrypoints/add_bundle_to_fbc.py
@@ -466,6 +466,13 @@ def release_bundle_to_fbc(args: argparse.Namespace, bundle: Bundle) -> None:
     for release_config_template in bundle.release_config["catalog_templates"]:
         template_name = release_config_template["template_name"]
         template_mapping = get_catalog_mapping(bundle.operator.config, template_name)
+        default_channel = release_config_template.get("defaultChannel")
+        template_channels = release_config_template["channels"]
+        if default_channel and default_channel not in template_channels:
+            raise ValueError(
+                f"defaultChannel '{default_channel}' not found in '{template_name}' "
+                f"channels: {template_channels}"
+            )
         if not template_mapping:
             raise ValueError(
                 f"Template mapping not found for '{template_name}' in ci.yaml."
--- a/operatorcert/entrypoints/add_bundle_to_fbc.py
+++ b/operatorcert/entrypoints/add_bundle_to_fbc.py
@@ -299,7 +299,7 @@ class BasicTemplate(CatalogTemplate):
         package = {
             "schema": "olm.package",
             "name": self.operator.operator_name,
-            "defaultChannel": channels[0]["name"],
+            "defaultChannel": release_config.get("defaultChannel", channels[0]["name"]),
         }
         bundle_obj = {"schema": "olm.bundle", "image": bundle_pullspec}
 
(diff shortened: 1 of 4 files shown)


acornett21:
@haripate Can the `fbc_autorelease.md` also be updated so that partners know this is supported?

mantomas (approved):
Provided changes looks good.

operatorcert/entrypoints/add_bundle_to_fbc.py
```diff
@@ -466,6 +466,13 @@ def release_bundle_to_fbc(args: argparse.Namespace, bundle: Bundle) -> None:
     for release_config_template in bundle.release_config["catalog_templates"]:
         template_name = release_config_template["template_name"]
         template_mapping = get_catalog_mapping(bundle.operator.config, template_name)
+        default_channel = release_config_template.get("defaultChannel")
+        template_channels = release_config_template["channels"]
+        if default_channel and default_channel not in template_channels:
+            raise ValueError(
+                f"defaultChannel '{default_channel}' not found in '{template_name}' "
+                f"channels: {template_channels}"
```
BorekZnovustvoritel: Suggestion: this could be slightly reworded for clarity, something like
```py
f"defaultChannel '{default_channel}' not found in '{template_name}'. "
f"Available channels: {template_channels}"
```

tests/entrypoints/test_add_bundle_to_fbc.py
```diff
@@ -434,8 +436,25 @@ def test_release_bundle_to_fbc(
...
-            {"template_name": "fake-basic.yaml"},
-            {"template_name": "fake-semver.yaml"},
+            {
+                "template_name": "fake-template.yaml",
+                "channels": ["alpha"],
+                "defaultChannel": "stable",
+            },
+        ]
+    }
+    mock_catalog_mapping.side_effect = {
```
BorekZnovustvoritel: I don't think this does what you expect it to do. From the [docs](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.side_effect) it seems like if you pass an iterable (dict is iterable) to a side_effect, it returns the iterator-yielded values, which would be the dictionary keys in this case.
You can verify it by running this snippet:

```python
from unittest.mock import MagicMock
m = MagicMock()
m.side_effect = {"foo": 1, "bar": 2}
print(m()) # shows "foo"
print(m()) # shows "bar"
```

I guess this meant to specify the `return_value` instead.

BorekZnovustvoritel (commented):
One problem I spotted in the tests, otherwise this seems good to me.

operatorcert/entrypoints/add_bundle_to_fbc.py
haripate: Updated

tests/entrypoints/test_add_bundle_to_fbc.py
haripate: Thank you, @BorekZnovustvoritel for catching this. Yes I missed to add the value as list instead of dict there. Updated.

haripate:
@acornett21 I have also updated the `fbc_autorelease.md` along with the changes in this PR.

BorekZnovustvoritel (approved):

PR #940, opened 2026-04-28
[ISV-7093] guard hosted merge on base branch change

mantomas:
Closes ISV-7093

Handles situation when multiple parallel PRs (and pipelines) are touching the same operator or catalog and merging one of them can make the other PR(s) invalid (e.g. multiple channel heads).

The solution creates fingerprint (hash) of the updated operator/catalog tree-ish in a new task right after `detect-changes` and is then used as a guard in `merge-pr`:
- uses target branch tip (or reuse clone-repository-base if OID match) to create fingerprint
- pass if the target tip oid does not change (nothing was merged during the pipeline run)
- calculate and compare new fingerprint if base tip moved
 - - pass if fingerprint match (PR related catalog or operator is unchanged)
 - - guards the merge on fingerprint mismatch

Known limitations: as a naive solution, there is no actual `opm validate`, leaving some space for false positives (stated in the PR comment if guarded, pipeline re-run should fix it). This trade off is time effective, regular validation with catalog/index build will introduce another time-window for race condition.

[Example guarded merge is here.](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/343#issuecomment-4342927405)
- fixed by [pipeline rerun](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/343#issuecomment-4343249357) as there was no actual conflict in this case

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests


ansible/roles/operator-pipeline/templates/openshift/tasks/record-merge-base-lane-snapshot.yml
```diff
@@ -0,0 +1,144 @@
...
+          TIP_OID="${TIP_OID_REMOTE}"
+          # safe.directory for base workspace (merge-base-lane-fingerprint); env only, no $HOME writes.
+          export GIT_CONFIG_VALUE_0="${TIP_CLONE}"
+        else
+          rm -rf "${TIP_CLONE}"
+          # safe.directory before git -C on the new clone (ownership vs step user).
+          export GIT_CONFIG_VALUE_0="${TIP_CLONE}"
+          git clone --depth 1 -b "$(params.git_base_branch)" "${REMOTE_URL}" "${TIP_CLONE}"
+          TIP_OID="$(git -C "${TIP_CLONE}" rev-parse HEAD)"
+        fi
```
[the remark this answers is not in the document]
mantomas: The snapshot step always resolves the live refs/heads/<git_base_branch> tip via ls-remote and records base_oid from that tip (reusing the base workspace only when that OID equals git_commit_base; otherwise it shallow-clones that tip and fingerprints HEAD). merge-base-lane-fingerprint verify compares merge-time ls-remote to base_oid from the snapshot, so drift after the snapshot is detected relative to the tip we actually fingerprinted, including the case where the branch was already ahead of git_commit_base before the snapshot ran.
What stays at git_commit_base is detect-changes and the base workspace used there—that’s the intentional scope called out in the README: lane-level fingerprinting, not a second full validation pass on the newer tip.

bclindner (approved):
thanks for the README add, was having trouble making sense of this - my understanding of the logic's still a little fuzzy but I think this looks good

tomasbakk:
I have no objections/nitpicks. Overall it LGTM :+1:

tomasbakk (approved):

ansible/roles/operator-pipeline/templates/openshift/tasks/merge-pr.yml
```diff
@@ -19,10 +19,37 @@ spec:
...
       description: Scratch space and storage for the comment and related data
+    - name: merge-base-guard
+      description: Snapshot from record-merge-base-lane-snapshot (snapshot.json).
+      optional: true
+    - name: ssh-directory
+      optional: true
+      description: >
+        Same as git-clone: .ssh with known_hosts and private key. Required for
+        merge-base-lane-fingerprint verify when git_repo_url uses SSH.
+    - name: basic-auth
```
Allda: I don't see this workspace being used anywhere? Do we need it? If not, can it be removed to reduce the complexity of the code?

mantomas: I see now, thanks for pointing out! I was concerned about matching the logic with git-clone and didn't realized this optional workspace is not used in hosted pipeline - I will remove it from the task.

ansible/roles/operator-pipeline/templates/openshift/tasks/merge-pr.yml
```diff
@@ -64,12 +101,56 @@ spec:
...
+            export GIT_CONFIG_VALUE_0="$(workspaces.source.path)"
+            VERIFY_RC=0
+            merge-base-lane-fingerprint verify \
+              --git-dir "$(workspaces.source.path)" \
+              --git-repo-url "$(params.git_repo_url)" \
+              --git-base-branch "$(params.git_base_branch)" \
+              --snapshot-file "$(workspaces.merge-base-guard.path)/snapshot.json" \
+              || VERIFY_RC=$?
+            if [[ "$VERIFY_RC" -eq 10 ]]; then
+              MSG="The target branch moved after this run recorded its merge-guard snapshot, "
```
Allda: Can we simplify the messaging here? When reading it as someone who didn't implement the logic, I am quite confused about what this means. Especially the parts `recorded its merge-guard snapshot` or `in the guard lane changed` are hard to understand. Please exclude any internal logic or jargon from the message and make it more user-friendly.

ansible/roles/operator-pipeline/templates/openshift/tasks/record-merge-base-lane-snapshot.yml
```diff
@@ -0,0 +1,144 @@
...
+      description: >
+        Clone of the repo at git_commit_base (detect-changes). Reused for fingerprinting
+        when the live remote tip still matches git_commit_base (no second clone).
+    - name: output
+      description: Writes snapshot.json and ephemeral tip clone (e.g. results/merge-base-guard).
+    - name: ssh-directory
+      optional: true
+      description: >
+        Same as git-clone / merge-pr. Required for SSH git_repo_url when cloning target tip.
+    - name: basic-auth
```
Allda: Same here, this is not used.

operatorcert/entrypoints/merge_base_lane_fingerprint.py
```diff
@@ -0,0 +1,118 @@
...
+        "record", help="Write snapshot.json from the base clone at tree-ish"
+    )
+    rec.add_argument(
+        "--git-dir",
+        required=True,
+        type=Path,
+        help="Path to git repository (base branch clone)",
+    )
+    rec.add_argument(
+        "--tree-ish",
```
Allda, line 41: What does this mean? Is it a Git revision?

mantomas: The [terminology](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-tree-ishalsotreeish) overlaps here, but it is directly passed to the `git ls-tree <tree-ish>` so it matches the git [documentation](https://git-scm.com/docs/git-ls-tree#_synopsis) for `ls-tree` and we are then fingerprinting the tree structure for the given path, so in my POV it matches the intention. We can rename this to `revision` if needed to make the terminology closer to pipeline.

ansible/roles/operator-pipeline/templates/openshift/tasks/merge-pr.yml
mantomas: Message changed, [here is the output](https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/344#issuecomment-4379525822) - does it make sense this way? I'm biased so might not see the issue clearly.

ansible/roles/operator-pipeline/templates/openshift/tasks/merge-pr.yml
mantomas: basic-auth removed

ansible/roles/operator-pipeline/templates/openshift/tasks/record-merge-base-lane-snapshot.yml
mantomas: basic-auth removed

Allda:
@mantomas, it looks like you started using a different docstring format than the rest of the codebase. Any reason for switching from Google docstring format?

tests/operatorcert/test_merge_base_lane.py
```diff
@@ -0,0 +1,421 @@
...
+
+
+@patch("operatorcert.merge_base_lane.subprocess.run")
+def test_git_ls_tree_lines_git_fails(mock_run: MagicMock, tmp_path: Path) -> None:
+    mock_run.return_value = MagicMock(returncode=128, stderr="fatal", stdout="")
+    with pytest.raises(RuntimeError, match="git ls-tree"):
+        git_ls_tree_lines(tmp_path, "nope", ["x"])
+
+
+def test_git_ls_tree_lines_sorted(tmp_path: Path) -> None:
```
Allda: I was looking for a test that verifies a happy path of this function, and the only test I was able to find is this one, and from my understanding, this test only tests the `sorted` function, but doesn't test the inner logic. So if there is a bug in the code and the function returns `[]`, this test always passes. Is this tested somewhere else?

ansible/roles/operator-pipeline/templates/openshift/tasks/record-merge-base-lane-snapshot.yml
```diff
@@ -0,0 +1,131 @@
...
+          rm -rf "${TIP_CLONE}"
+          TIP_CLONE="${BASE_CLONE}"
+          TIP_OID="${TIP_OID_REMOTE}"
+          # safe.directory for base workspace (merge-base-lane-fingerprint); env only, no $HOME writes.
+          export GIT_CONFIG_VALUE_0="${TIP_CLONE}"
+        else
+          rm -rf "${TIP_CLONE}"
+          # safe.directory before git -C on the new clone (ownership vs step user).
+          export GIT_CONFIG_VALUE_0="${TIP_CLONE}"
+          git clone --depth 1 -b "$(params.git_base_branch)" "${REMOTE_URL}" "${TIP_CLONE}"
```
Allda, line 122: Can this have any impact on the current storage limit? Does it double the required size?

Allda:
Overall, I think I am ok with the PR based on what I have seen in the examples and tests. Please update the dostrings to align with the rest of the codebase, and if there is any room for improving unit tests to cover the functionality rather than just coverage please add it. I'll leave it up to you.

ansible/roles/operator-pipeline/templates/openshift/tasks/record-merge-base-lane-snapshot.yml
mantomas: This is actually a very good question. There are two possible clone added and in the current implementation it will require extending the storage size. Both of the clones are needed (cannot do ls-tree inspection against remote), but are only relevant to a single task. I will change this to use scratch emptyDir instead in both places to completely remove the workspace dependency - thanks for pointing out!

tests/operatorcert/test_merge_base_lane.py
mantomas: Good catch - updated the tests to be exact in the assertion. I mostly verified the underlying logic by running the pipelines, but this should be definitely covered also in tests.

ansible/roles/operator-pipeline/templates/openshift/tasks/record-merge-base-lane-snapshot.yml
mantomas: Updated to use emptyDir for temporary clones, so workspace size change is not needed. Verified in separate pipeline for both collision and happy path: https://github.com/redhat-openshift-ecosystem/community-operators-pipeline-preprod/pull/345

mantomas:
Docstrings updated to match the rest of the codebase.

PR #973, opened 2026-06-23
adding log message for projects that chose undistributed, so it's cle…

acornett21:
…arer when their operator isn't released to catalogs

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

## Motivation
PR's like the below, there are no logs explaining why `acquire-lease` is skipped and only a bundle is released. This adds a finally block so this is always logged for clarity, and does not affect any of the workflow revolving around the lease itself.

- Relates: redhat-openshift-ecosystem/certified-operators#8761

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
@@ -882,6 +882,8 @@ spec:
           value: "$(context.pipeline.name)"
         - name: pipeline_status
           value: "$(tasks.status)"
+        - name: operator_distribution
+          value: "$(tasks.get-pyxis-certification-data.results.operator_distribution)"
       workspaces:
         - name: base
           workspace: repository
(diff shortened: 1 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-release-pipeline.yml
```diff
@@ -685,6 +685,31 @@ spec:
 
   finally:
 
+    # Log when acquire lease was skipped because the partner chose 'undistributed' in pyxis
+    - name: log-distribution-skip-notice
```
Allda: A better solution would be to extend the `github-add-summary-comment` and append the user-facing comment and method that the selected distribution method didn't ship the bundle to the catalog. This is more visible than a log message in the log file.

acornett21: I agree with that. Thanks.

ansible/roles/operator-pipeline/templates/openshift/tasks/github-pipelinerun-summary.yml
```diff
@@ -53,6 +53,11 @@ spec:
         A flag that determines whether pipeline logs should be uploaded to Github gists
       type: string
 
+    - name: operator_distribution
```
Allda, line 56: The task is also used in the hosted pipeline, and since this task is required, it will break it. Adding a default value to this parameter with `""` should fix it, as we don't need this value in the hosted pipeline.

Allda (approved):

PR #987, opened 2026-07-08
feat(ISV-7356): certProject check fallback to base branch on operator removal

mantomas:
- moved the cert-project-check to separate Python module
- fallback to use base branch `ci.yaml` when operator is deleted in the PR
- fix invalid reference in check-permissions `deleted_catalog_operators` vs `removed_catalog_operators` 
- cosmetic changes to `tox.ini` and docs as I was unable run it locally after the change to Poetry

Closes ISV-7356

Example ISV operator remove PR is here: https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/3711#issuecomment-4912866317

AI-assisted-by: Cursor

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests


operatorcert/entrypoints/certification_project_check.py
```diff
@@ -0,0 +1,194 @@
...
+    parser.add_argument(
+        "--affected-catalog-operators",
+        default=[],
+        action=SplitArgs,
+        help="Comma separated list of affected catalog operators",
+    )
+    parser.add_argument(
+        "--cert-project-required",
+        default="true",
+        help="Whether a cert project ID must be present",
```
kosciCZ, line 51: Nitpick: this should probably be just a flag (either specified or not). with action='store_true'

operatorcert/entrypoints/certification_project_check.py
```diff
@@ -0,0 +1,194 @@
...
+    try:
+        cert_project_id = check_certification_projects(
+            args.repo_head_path,
+            args.repo_base_path,
+            operator_names,
+        )
+    except (FileNotFoundError, ValueError) as exc:
+        LOGGER.error("%s", exc)
+        sys.exit(1)
+    print(cert_project_id)
```
kosciCZ: Should this be a LOGGER.info() ?

operatorcert/entrypoints/certification_project_check.py
```diff
@@ -0,0 +1,194 @@
...
+    Build a set of operator names affected by the pull request.
+
+    Args:
+        affected_operators (list[str]): Operator names from detect-changes
+        affected_catalog_operators (list[str]): Catalog operators in catalog/version format
+
+    Returns:
+        set[str]: Unique affected operator names
+    """
+    operators = {op.strip() for op in affected_operators if op and op.strip()}
```
kosciCZ: Is there a need to do stripping here if the SplitArgs was modified to do the same thing?

tox.ini
```diff
@@ -59,29 +74,32 @@ commands =
...
     # pip-audit does not support not failing on unfixable vulnerabilities, this hacks around that
-    python ./local-dev/pip-audit-parse.py  /tmp/audit-output.json
+    poetry run python ./local-dev/pip-audit-parse.py  /tmp/audit-output.json
 
 
 [testenv:hadolint]
 allowlist_externals = hadolint
+depends =
 deps =
 commands_pre =
```
kosciCZ, line 96: This has a bunch of empty vars, is that intended?

kosciCZ (changes requested):

operatorcert/entrypoints/certification_project_check.py
mantomas: It is an existing pipeline parameter (true, false strings) - changing this to flag would require translate it in the task instead of just passing, creating additional Bash logic.

tox.ini
mantomas: This is the way how to override global behavior locally. Hadolint does not need any of these (it is part of the tox image) - it looks verbose but it actually makes it faster and resource light.

operatorcert/entrypoints/certification_project_check.py
mantomas: You are right - after handling Qodo review this logic is in Argparse action now. Removed the strip and fixed the tests. I kept the `strip` down in the `split` section as there are theoretical edge cases not handled by the SplitArgs.

operatorcert/entrypoints/certification_project_check.py
mantomas: I changed this to log.info, it matches the other Python modules now. Note: for me, looking at the gists and its format, to match the Bash `echo` simple `print` statement would be cleaner as we are storing any output - the log is adding additional formatting making the gist more chaotic (not mentioning the _double_ formatted Splunk logs).

tox.ini
kosciCZ: Very cool, I did not know that. Could maybe use a comment, but not a blocker

operatorcert/entrypoints/certification_project_check.py
kosciCZ: This might be a potential improvement for the entire pipeline, to rethink how we format logs. Thank you for unifying it in the script

operatorcert/entrypoints/certification_project_check.py
kosciCZ: Makes sense unfortunately, let's keep it as is then

kosciCZ (approved):
LGTM, thanks for addressing my comments!

tomasbakk:
Please forgive me for reacting to this later, I went through few days ago it but forgot to put anything here

tomasbakk (approved):

PR #988, opened 2026-07-09
removing humanize.naturaldelta() for task output duration

acornett21:
...to ensure long-running tasks no longer output "an hour" when they actually run longer then an hour. This should prevent requests to increase timeouts past 1hr, which has already been done.

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

- Relates: redhat-openshift-ecosystem/community-operators-prod#10312


operatorcert/tekton.py
```diff
@@ -4,7 +4,6 @@
 import json
 from typing import Any, Dict
 
-import humanize
```
Allda, line 7: If you want this to be removed, please also remove the dependency from the Python package using poetry. This is the only place where this library is used.

acornett21: @Allda I've updated the dependencies. Do you think change and display format makes more sense, and is easier to troubleshoot issues?

Allda: I guess this change seems fine if it helps you to clarify things.

Allda (approved):

operatorcert/tekton.py
acornett21: I'm open to other ideas, if there are better libs, but logging `an hour` for things that are over that doesn't make sense to me. Then if something is 1.5 hours would get logged as `two hours`, which seems even more confusing.

jfrancin: For development and build environments (which is what GitHub is), 'humanizing' times _anywhere_ is a bad idea. GitHub is already reticent about using explicit dates/times in other places - let's not help it hide info by using this package. I support this change.

PR #992, opened 2026-07-17
fix(ISV-7433): registry secret resolution after OPM bump

mantomas:
After OPM bump in https://github.com/redhat-openshift-ecosystem/operator-pipelines/pull/990 there are errors in stage environment like this https://github.com/redhat-openshift-ecosystem/certified-operators-preprod/pull/7473#issuecomment-4998878082
The previous version of OPM was 1.46.0 and the new is 1.71.0 - in 1.53.0 new secret resolution was introduced https://github.com/operator-framework/operator-registry/releases/tag/v1.53.0 as a possible breaking change. OPM now expects `config.json` and our defined `.dockerconfigjson` is picked up only if the first does not exist. If the file exist but it is empty, then it is silently passed as the secret.
This PR copies the `.dockerconfigjson` to the right location. Verified in this PR https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/3773 against the current code as our integration test do not cover this workflow.

Assisted-by: Cursor

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/tasks/add-bundle-to-fbc.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/tasks/add-bundle-to-fbc.yml
@@ -65,13 +65,19 @@ spec:
     - name: add-bundle-to-fbc
       image: "$(params.pipeline_image)"
       workingDir: $(workspaces.source.path)
-      env:
-        # The registry auth file for opm render catalog command
-        - name: REGISTRY_AUTH_FILE
-          value: "$(workspaces.registry-credentials.path)/.dockerconfigjson"
       script: |
         #! /usr/bin/env bash
         set -xe
+
+        # OPM 1.53+ (containers/image) expects a Docker-style auth layout:
+        # DOCKER_CONFIG/config.json and/or REGISTRY_AUTH_FILE pointing at that file.
+        # K8s dockerconfigjson secrets mount as .dockerconfigjson; reshape for OPM.
+        if [[ "$(workspaces.registry-credentials.bound)" == "true" ]]; then
+          export DOCKER_CONFIG="$(mktemp -d)"
+          cp "$(workspaces.registry-credentials.path)/.dockerconfigjson" \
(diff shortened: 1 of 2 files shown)


ansible/roles/operator-pipeline/templates/openshift/tasks/add-bundle-to-fbc.yml
```diff
@@ -65,14 +65,20 @@ spec:
     - name: add-bundle-to-fbc
       image: "$(params.pipeline_image)"
       workingDir: $(workspaces.source.path)
-      env:
-        # The registry auth file for opm render catalog command
-        - name: REGISTRY_AUTH_FILE
```
BorekZnovustvoritel, line 70: Using search, I found out this is not the only usage of this variable.

It is also referenced in tasks:
* `verify-index-signatures` - it seems like there it is used by Podman, no action required
* `digest-pinning` - looks like this passed even in the failed pipeline, no action required
* `build-fbc-scratch-catalog`, where OPM **is** used
* some tests

I see that even the `build-fbc-scratch-catalog` passed, but I am not sure if this is a coincidence or if some part of `opm` does not use that breaking change yet. Can you please take a look and chime in to the discussion if we need to do the same process for that task too?

mantomas: Thanks for thorough insight. I already checked all the places and kept the fix minimal to where it is needed - it makes sense to keep the same fix in all places for the future. 
If I'm not mistaken, the pull in `build-fbc-scratch-catalog` is passing simply because the image in Quay is public (and no pull from Connect registry at all). Also I'm not sure if there might be situations when the pull needs actual authentication so I will keep the secret with the path reshape.
Good call!

BorekZnovustvoritel (approved):

PR #996, opened 2026-07-21
ISV-7466 Separate queue for ISV and community

Allda:
The dispatcher capacity manager previously counted all running pipelines by name only. Since ISV, marketplace, and
community operators all run under the same pipeline name, there was no way to enforce separate capacity limits per
repository.

This PR fixes that by:
1. Stamping each PipelineRun with a operator-pipelines.redhat.com/repository annotation (taken from the webhook
payload) when it's triggered
2. Teaching the capacity manager to filter by that annotation, so each repository's running pipelines are counted
independently

As a result, a burst of community PRs can no longer consume capacity reserved for certified operators.

### DIFF
--- a/ansible/inventory/group_vars/operator-pipeline-integration-tests-community.yml
+++ b/ansible/inventory/group_vars/operator-pipeline-integration-tests-community.yml
@@ -17,6 +17,8 @@ operator_pipeline_dispatcher_config:
       pipeline_name: "operator-hosted-pipeline"
       max_capacity: "{{ operator_pipeline_dispatcher_hosted_capacity }}"
       namespace: "{{ oc_namespace }}"
+      annotation_selector:
+        operator-pipelines.redhat.com/repository: "{{ integration_tests_git_upstream_repo }}"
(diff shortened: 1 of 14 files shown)


wcheang (approved):

kosciCZ (approved):
LGTM, this is a nice improvement. @JakubDurkac also had a suggestion that it might be worth to limit release pipelines to 2 because most of the time is spent on acquiring lease, and then on the actual IIB build. Starting multiple pipelines in parallel makes them wait anyway on the acquire-lease step, which is not very efficient and doesn't provide any real speed-up, because the part before the acquire-lease takes just a few minutes. 

Not sure if this concern might become invalid with the new IIB achitecture that allows for faster builds(?). Anyway, something to consider, but not a blocker for this PR

JakubDurkac:
Yeah, and btw, I’ve added a comment to the jira ISV-7466 which explains what that suggestion might fix and why. You can check it and see if it’s relevant or not if you want to.

PR #997, opened 2026-07-21
fix(ISV-7482): update git-init image from RHEL8 to RHEL9

kosciCZ:
The git-init image suddenly stopped working. This caused clone-repository tasks to fail with "Permission denied (publickey)".

Upon further investigation, this seems to boil down to OpenSSH version. The root cause seems to be that GitHub no longer accepts SSH auth from OpenSSH 8.0 (RHEL8) which lacks support for the publickey-hostbound-v00@openssh.com extension. This commit upgrades pipelines-git-init from RHEL8 (OpenSSH 8.0) to RHEL9 (OpenSSH 9.9) which supports the required extension. Even if that wasn't the root cause (no public GitHub announcement was found), this makes it work again.

Investigation was assisted by Claude Code (claude-opus-4-6)

### DIFF
--- a/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
+++ b/ansible/roles/operator-pipeline/templates/openshift/pipelines/operator-ci-pipeline.yml
@@ -15,7 +15,7 @@ spec:
     - name: gitInitImage
       description: The image providing the git-init binary that the checkout task runs.
       type: string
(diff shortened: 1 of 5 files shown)


mantomas (approved):

kosciCZ:
Note: related to this github incident https://www.githubstatus.com/incidents/g40zcbvchny4

Allda (approved):

PR #998, opened 2026-07-22
feat(ISV-7431): redact potential leaks in logs

BorekZnovustvoritel:
### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

# Testing of this
Deployed integration tests with dummy "leaks" from local host. Results:
* Latest version (after resolving Quodo comments: https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/3821), check the [published gist with log file](https://gist.github.com/rh-operator-bundle-bot/64568e9115d0fc2a602bfff8f1a42361#file-operator-hosted-pipeline-runx69tr-log-L5).
* Testing of dummy leak in static test results: https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/3820#issuecomment-5055850921

No actual secret was leaked, the "token" is 3 consecutive `{"hello": "world"}` encoded JSONs joined with dots.

## Shortcomings
As you can see, the redaction string is shifted by 1 byte. I have [reported this issue to upstrem](https://github.com/leaktk/leaktk/issues/280).

I also have an almost-finished version of my own redaction wrapper based on the report from `leaktk scan`, which I can update to handle this one-byte shift if necessary (but the whole wrapper is around 400 LoC and 500 lines of test code). I have this code because the `leaktk redact` command didn't exist when I started working on this PR.

The data uploaded to Pyxis is not modified in any way.

### DIFF
--- a/operatorcert/redact.py
+++ b/operatorcert/redact.py
@@ -7,9 +7,9 @@ and then redacts.
 """
 
 import logging
+import subprocess
 import tempfile
 from pathlib import Path
-from subprocess import Popen, PIPE
 import json
 
 from pydantic import BaseModel
@@ -47,16 +47,19 @@ def scan(*input_paths: Path) -> list[ResultSetModel]:
     Returns:
         Parsed scan results.
     """
-    results = []
-    for input_path in input_paths:
-        with Popen(
-            ["leaktk", "scan", "--kind", "Files", str(input_path.absolute())],
-            stdout=PIPE,
-        ) as scan_proc:
-            result_bytes = scan_proc.stdout.read()  # type: ignore[union-attr]
-        loaded_dict = json.loads(result_bytes)
-        results.append(ResultSetModel.model_validate(loaded_dict))
-    return results
+    requests = "".join(
+        json.dumps(
+            {"id": str(i), "kind": "Files", "resource": str(input_path.absolute())}
+        )
+        + "\n"
+        for i, input_path in enumerate(input_paths)
+    )
+    results_jsonl = subprocess.check_output(
+        ["leaktk", "listen"], input=requests, text=True
+    )
+    return list(
+        map(ResultSetModel.model_validate, map(json.loads, results_jsonl.splitlines()))
+    )
 
 
 def _redact(
@@ -78,20 +81,20 @@ def _redact(
             delete=False
         )
         with open(input_path, "rb") as input_file:
-            with Popen(
+            redact_proc = subprocess.run(
                 ["leaktk", "redact", "--kind", "Stdio"],
                 stdin=input_file,
                 stdout=tmp_file.file,
-                stderr=PIPE,
-            ) as redact_proc:
-                redact_proc.wait()
-                if redact_proc.returncode != 0:
-                    raise RuntimeError(
-                        f"Redact failed with return code: "
-                        f"{redact_proc.returncode}. STDERR: "
-                        f"{redact_proc.stderr.read().decode('utf-8')}"  # type: ignore[union-attr]
-                    )
-                file_mapping[input_path.absolute()] = Path(tmp_file.name).absolute()
+                stderr=subprocess.PIPE,
+                check=False,
+            )
+            if redact_proc.returncode != 0:
+                raise RuntimeError(
+                    f"Redact failed with return code: "
+                    f"{redact_proc.returncode}. STDERR: "
+                    f"{redact_proc.stderr.decode('utf-8')}"
+                )
+            file_mapping[input_path.absolute()] = Path(tmp_file.name).absolute()
         tmp_file.close()
     return file_mapping
 
--- a/operatorcert/redact.py
+++ b/operatorcert/redact.py
@@ -1,0 +1,132 @@
+"""
+Module for redacting leaked information from output.
+Uses LeakTK: https://github.com/leaktk/leaktk
+
+Parses LeakTK output, concatenates its results so no overlaps exist
+and then redacts.
+"""
+
+import logging
+import tempfile
+from pathlib import Path
+from subprocess import Popen, PIPE
+import json
+
+from pydantic import BaseModel
+
+LOGGER = logging.getLogger("operator-cert")
+
+
+class RedactLocation(BaseModel):
+    """
+    Class for tracking information about a chunk of data in files.
+    """
+
+    path: Path
+
+
+class ResultModel(BaseModel):
+    """
+    Model for parsing a single result from scanning.
+    """
+
+    location: RedactLocation
+
+
+class ResultSetModel(BaseModel):
+    """
+    Model for parsing all results from scanning.
+    """
+
+    results: list[ResultModel]
+
+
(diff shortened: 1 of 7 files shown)


tests/test_redact.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+def _scan_result_bytes(path: str) -> bytes:
+    return json.dumps({"results": [{"location": {"path": path}}]}).encode()
+
+
+def _make_popen_side_effect(
+    scan_result: bytes,
+    redact_returncode: int = 0,
+) -> Any:
```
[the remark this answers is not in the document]
BorekZnovustvoritel: I mean yeah, but this is just a side_effect function. Using broad type on this does not have so dire consequences.

operatorcert/redact.py
```diff
@@ -0,0 +1,129 @@
...
+            ) as redact_proc:
+                redact_proc.wait()
+                if redact_proc.returncode != 0:
+                    raise RuntimeError(
+                        f"Redact failed with return code: "
+                        f"{redact_proc.returncode}. STDERR: "
+                        f"{redact_proc.stderr.read().decode('utf-8')}"  # type: ignore[union-attr]
+                    )
+                file_mapping[input_path.absolute()] = Path(tmp_file.name).absolute()
+    return file_mapping
```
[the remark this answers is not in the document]
BorekZnovustvoritel: The file has to survive outside off the scope of this function. Updated the docstring to specify that a tempfile is created and added a cleanup to the gist creating function, where this is called at the moment.

operatorcert/redact.py
```diff
@@ -0,0 +1,132 @@
...
+def scan(*input_paths: Path) -> list[ResultSetModel]:
+    """
+    Scan all input paths and return a list of the results.
+    Returns:
+        Parsed scan results.
+    """
+    results = []
+    for input_path in input_paths:
+        with Popen(
+            ["leaktk", "scan", "--kind", "Files", str(input_path.absolute())],
```
mpapadopoullos: It would take me a while to review how the whole task works; so I could really use your knowledge to speed things up.

Could the value of input_path ever be manipulated? (e.g: somehow a file `test;whoami` reach the `Popen` call?

BorekZnovustvoritel: This is at this point called only for Gist creation. The gist file names are sourced either from pipelineRun name (generated by tekton) or hard-set to a value in the pipeline configuration.

On top of this, this uses Popen without `shell=True`, so `test;whoami` should stay as a string and `whoami` would not get executed. Tried on this example:

```python
>>> import subprocess
>>> a=subprocess.Popen(["echo", "foo; whoami"])
foo; whoami
```

mpapadopoullos: true!

mantomas (approved):
Is there any test pipeline where it actually ran against some secret or it is only unit-tested?
LGTM - the logic seems correct to me.

johnbe11 (approved):

operatorcert/redact.py
```diff
@@ -0,0 +1,132 @@
...
+            stdout=PIPE,
+        ) as scan_proc:
+            result_bytes = scan_proc.stdout.read()  # type: ignore[union-attr]
+        loaded_dict = json.loads(result_bytes)
+        results.append(ResultSetModel.model_validate(loaded_dict))
+    return results
```
bplaxco: Aside: If you want to do a bulk scan:

```python
requests = "".join(
    json.dumps({"id": str(i), "kind": "Files", "resource": input_path}) + "\n"
    for i, input_path in enumerate(input_paths)
)

with subprocess.Popen(
    ["leaktk", "listen"],
    text=True,
    stdin=subprocess.PIPE,
    stdout=subprocess.PIPE,
) as p:
    p.stdin.write(requests)
    p.stdin.close()
    results = list(map(ResultSetModel.model_validate, map(json.loads, p.stdout)))
```

Or if using check_output:

```python
requests = "".join(
    json.dumps({"id": str(i), "kind": "Files", "resource": input_path}) + "\n"
    for i, input_path in enumerate(input_paths)
)
results_jsonl = subprocess.check_output(["leaktk", "listen"], input=requests, text=True)
results = list(
    map(ResultSetModel.model_validate, map(json.loads, results_jsonl.splitlines()))
)

```

operatorcert/redact.py
```diff
@@ -0,0 +1,132 @@
...
+                stdout=tmp_file.file,
+                stderr=PIPE,
+            ) as redact_proc:
+                redact_proc.wait()
+                if redact_proc.returncode != 0:
+                    raise RuntimeError(
+                        f"Redact failed with return code: "
+                        f"{redact_proc.returncode}. STDERR: "
+                        f"{redact_proc.stderr.read().decode('utf-8')}"  # type: ignore[union-attr]
+                    )
```
bplaxco: Would subprocess.run be cleaner here?


```suggestion
            redact_proc = subprocess.run(
                ["leaktk", "redact", "--kind", "Stdio"],
                stdin=input_file,
                stdout=tmp_file.file,
                stderr=PIPE,
            )
            if redact_proc.returncode != 0:
                raise RuntimeError(
                    f"Redact failed with return code: "
                    f"{redact_proc.returncode}. STDERR: "
                    f"{redact_proc.stderr.decode('utf-8')}"  # type: ignore[union-attr]
                )
```

operatorcert/redact.py
```diff
@@ -0,0 +1,132 @@
...
+            ["leaktk", "scan", "--kind", "Files", str(input_path.absolute())],
+            stdout=PIPE,
+        ) as scan_proc:
+            result_bytes = scan_proc.stdout.read()  # type: ignore[union-attr]
+        loaded_dict = json.loads(result_bytes)
+        results.append(ResultSetModel.model_validate(loaded_dict))
+    return results
+
+
+def _redact(
```
bplaxco, line 65: When we add `--kind Files` for the redact command in LeakTK, will you want/need a flag that backs up the original file?

I'm thinking the default will be redact in place effectively so that there's less of a risk of the file being accidentally added during uploads and such. But I'm thinking something like `--backup-dir` and if it's set it'll move the original file there and put the redacted file where the original was.

Thoughts?

bplaxco (commented):
Redaction bits look good. Added some comments as potential tweaks to how some sub processes are ran but those are more nitpicks and optional changes.

Also looking forward to getting `leaktk redact --kind Files` out. If you have any requirements or requests for that one, please add them to: https://github.com/leaktk/leaktk/issues/78

operatorcert/redact.py
BorekZnovustvoritel: Some of the files are mounted in read-only state in the pipeline, so some of the files can't be overwritten. I would like an optional flag that would create redacted copies in the specified location.

Also I would like to have some output with "leaks found in files", preferably in JSON output, so that it can be used for alerting and finding the redacted files easily, if possible.

PR #999, opened 2026-07-23
feat(ISV-7470): update error message for check_replaces_availability

JakubDurkac:
Update error message to advice user to keep their fork up to date before each submission to include all versions merged in origin repo.

I've checked the other static tests to see if their error messages could be considered unclear in the same way, resulting in similar issues being filed, but it seems to me this is the only static test falling into that exact category. 

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/operatorcert/static_tests/common/bundle.py
+++ b/operatorcert/static_tests/common/bundle.py
@@ -263,7 +263,11 @@ def check_replaces_availability(bundle: Bundle) -> Iterator[CheckResult]:
     if replaces_version not in ver_to_dir:
         yield Fail(
             f"{bundle} attempts to replace version '{replaces_version}' which"
-            f" does not exist. Available versions: {sorted(ver_to_dir.keys())}"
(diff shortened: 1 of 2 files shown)


bclindner (approved):
very straightforward, looks good. i guess you could get rid of the f-strings anywhere we're not doing actual substitution but i kinda like that for cleanliness

JakubDurkac:
> very straightforward, looks good. i guess you could get rid of the f-strings anywhere we're not doing actual substitution but i kinda like that for cleanliness

I guess I'll just merge it as is for the sake of speed and try not to overthink on which alternative is nicer in the end. You're completely right that those could be omitted. My brain just automatically went for the same pattern as the previous lines :D. Some linters I think enforce that rule of "can't have f strings without the usage of formatting in them", luckily not in this project.

PR #1011, opened 2026-08-04
feat(ISV-7155): Use IIB overwrite feature in pipelines

JakubDurkac:
Draft: Can't be merged before IIB fixes the overwrite token in prod. Can be reviewed and tested via integration tests as the overwrite token fix is in IIB stage which is used by integration tests.

Changes:
- Reverted commit d87bc949b14a71c5dec7dd3adc67059a9dc5e872 that adds a workaround for passing overwrite token to IIB.
- Removed IIB call parameter build_tags as it's confirmed to exist for internal purposes only and doesn't create given tags for us.
- The normal tag (e.g. v4.23) is already created because of overwrite_from_index=True and using from_index version.
- The permanent tag (e.g. v4.23-17123456) is now created on our side, similarly to how it was done in the workaround before, but I switched to using authfile just as a safeguard from leaking the token in logs or exceptions propagation in python.

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/operatorcert/utils.py
+++ b/operatorcert/utils.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import re
 import subprocess
+import tempfile
 from typing import Any, Dict, List, Optional, Tuple
 
 import requests
@@ -15,6 +16,8 @@ from packaging.version import Version
 from requests import Session
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
+
+from operatorcert.integration.external_tools import RegistryAuthMixin
 
 LOGGER = logging.getLogger("operator-cert")
 
@@ -235,8 +238,6 @@ def copy_images_to_destination(
         cmd = [
             "skopeo",
             "copy",
-            "--retry-times",
-            "5",
             f"docker://{response.get('index_image_resolved')}",
             f"docker://{destination}:{version}{tag_suffix}",
         ]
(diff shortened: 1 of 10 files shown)


operatorcert/utils.py
```diff
@@ -248,6 +249,57 @@ def copy_images_to_destination(
...
+    try:
+        for response in iib_responses:
+            from_index = response.get("from_index", "")
+            src_image = response.get("index_image_resolved", "")
+            dest_image = f"{from_index}-{build_tags_suffix}"
+
+            cmd = [
+                "skopeo",
+                "copy",
+                "--format",
```
Allda: I would skip this option; it should not be needed.

JakubDurkac: Removed it.

JakubDurkac:
Next week till Wednesday, IIB is planning to have the blocker fix deployed to prod (https://github.com/release-engineering/iib/pull/1370). After that, we need to merge this PR, so IIB can switch to a new architecture on IIB stage and it won't affect our integration tests and preprod operator repos which use IIB stage, currently without the overwrite_token. Therefore, please, could you guys review this so it can be potentially deployed to operator-pipelines stage on Wednesday, Sep 16 (if everything is ready) and then to prod on Monday, Sep 21? Ales checked it previously and integration tests passed, but want to also have awareness and official reviews from other people in the team. @RichardPlesnik @mantomas 

I plan to merge this only once I get a ping from IIB that their fix is in prod and working, but I would like to have everything ready on our side before that.

mantomas (approved):
LGTM

JakubDurkac:
IIB has deployed the blocker fix to stage. I believe it shouldn’t affect us, but restarting tests just to be sure.

RichardPlesnik:
LGTM :+1:

RichardPlesnik (approved):

JakubDurkac:
Nevermind, the certified integration tests is failing on add-bundle-to-index - checking that out with IIB. And community is failing on hosted pipeline's preflight but so far 2 times at different spot.

PR #1020, opened 2026-08-18
fix(ISV-7499): remove cluster-admin from operator-pipeline-service-account

bclindner:
See title. `pipelines-custom-scc` seems to handle everything that actually does
need privilege escalation in this system - everything passes tests just fine
without the CRB reported as an issue here(?!). I may need a second look
by someone with more intuition on operator-pipelines to be sure here...

### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/ansible/roles/operator-pipeline/tasks/operator-pipeline-service-account.yml
+++ b/ansible/roles/operator-pipeline/tasks/operator-pipeline-service-account.yml
@@ -23,25 +23,3 @@
       kind: ServiceAccount
       metadata:
         name: "{{ service_account_name }}"
-
-- name: Grant SA role "{{ service_account_name }}"
-  tags:
-    - service-account
-    - init
-  kubernetes.core.k8s:
-    state: present
-    namespace: "{{ oc_namespace }}"
-    definition:
-      apiVersion: rbac.authorization.k8s.io/v1
-      kind: ClusterRoleBinding
-      metadata:
-        name: "{{ service_account_name }}-{{ oc_namespace }}-{{ item }}"
-      roleRef:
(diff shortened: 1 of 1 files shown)


bclindner:
@tomasbakk @wcheang if you can think of anything wrong w/ this let me know, ITs passed, but I'm a little hesitant b/c this feels like it was done for a reason, but the reason listed in the commit that added this doesn't seem to actually require cluster admin for this particular SA

wcheang:
Could you please double check with Ales, since he's the one that set up the SCC? According to the changes in the README when the SCC was set up, it's linked to the service account and requires the SA to have cluster admin access: https://github.com/redhat-openshift-ecosystem/operator-pipelines/commit/770dcafa5c371b45dea11fd1e922d151d42af83f

bclindner:
@Allda are you able to assist (provided you're still available for our team?)

Allda:
@bclindner I don't really remember why we added this. It was a long time ago. The successful run of integration tests should be a good enough indicator that all should be alright. However, when looking at the removed code it looks like it is part of a bash script that sets it once and it is not included in CI nor integration tests. So removing it here doesn't remove the SA from the cluster. Can you remove it manually from stage cluster and try to run tests again?

bclindner:
Tests seem to be passing after CRB removal in my stage cluster, I think this is good to go cc @Allda @wcheang @tomasbakk

wcheang (approved):

tomasbakk (approved):

PR #1029, opened 2026-08-31
docs(ISV-7458): add governance

BorekZnovustvoritel:
### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/governance.md
+++ b/governance.md
@@ -1,0 +1,117 @@
+# Project Governance
+
+# Project Overview
+
+The complete overview of this project is located in the README.md file in this
+repository.
+
+# Roles and Responsibilities
+
+## Maintainers
+
+Maintainers drive development and review contributions. They are responsible
+for decision-making and steering the development of this project. Work
+delivered by the maintainers should have a JIRA ticket before it is made.
+Larger changes should be discussed with the maintainer team in the form of
+a design doc that is then jointly refined in order to reach required
+functionality in a way that works well with the rest of the codebase and is
+maintainable.
+
+## Contributors
+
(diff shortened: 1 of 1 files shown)


governance.md
```diff
@@ -0,0 +1,117 @@
...
+
+* Any AI usage in the process of creating the code submitted **must** be
+  disclosed (see [Commit Standards](#commit-standards)).
+* The human submitting the changes **must** first review any AI output, fully
+  understand the code submitted and be able to explain the choices made during
+  the implementation.
+* If the effort put into creating the submitted code is less than the effort
+  required to give a thorough review, such a contribution is not valuable
+  enough to spend precious reviewer time on. This is especially true for
+  external contributions.
```
mantomas, line 71: This rule seems incomplete to me. It starts with `if` but I don't see any `then`, so what is the rule? It might be just my interpretation though.

BorekZnovustvoritel: `if effort_coding < effort_review: not_valuable_enough` is the idea here 😄 So this implies the PR will be most likely closed without merging. This point is copy-pasted from the internal contributing guidelines.

mantomas: I see - still does not make sense to me as a rule 😂 Maybe preceding it with that intent like `PR will be most likely closed without merging if ...`. Somehow it slipped my attention in previous reviews. Disregard my comment then if it makes sense to you as it is.

mantomas (approved):

PR #1043, opened 2026-09-09
feat(ISV-7663): operator name reservation depends on source

BorekZnovustvoritel:
### Merge Request Checklists

    - [ ] Linting
    - [ ] Code formatter - Black
    - [ ] Security scanners
    - [ ] Unit tests
    - [ ] Integration tests

### DIFF
--- a/operatorcert/entrypoints/reserve_operator_name.py
+++ b/operatorcert/entrypoints/reserve_operator_name.py
@@ -43,16 +43,46 @@ def setup_argparser() -> argparse.ArgumentParser:  # pragma: no cover
     return parser
 
 
-def check_operator_name_registered_for_association(args: Any) -> None:
+def has_approved_duplicate_name(args: Any) -> bool:
+    """
+    Check if duplicate name exception is allowed for this operator.
+    Args:
+        args: The parsed argparse arguments
+
+    Returns:
+        True if the exception is allowed, False otherwise.
+    """
+    if args.approved_exceptions:
+        for exception in set(args.approved_exceptions):
+            try:
+                match WorkflowException(exception):
+                    case WorkflowException.DUPLICATE_NAME:
+                        LOGGER.info(
+                            "Operator name is only checked against "
+                            "a single catalog source."
+                        )
+                        return True
+            except ValueError:
+                LOGGER.error("Unknown workflow exception %s is ignored", exception)
+    return False
+
+
+def check_operator_name_registered_for_association(
+    args: Any, allow_duplicates: bool = False
+) -> None:
     """
     Check if for given association/isv_pid operatorPackage already exist
     and validates that the package name match operator name requested.
     """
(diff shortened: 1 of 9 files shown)


operatorcert/entrypoints/reserve_operator_name.py
```diff
@@ -44,7 +44,8 @@ def check_operator_name_registered_for_association(args: Any) -> None:
         urljoin(
             args.pyxis_url,
             "v1/operators/packages?"
-            f"filter=association=={args.association};deleted!=true",
+            f"filter=association=={args.association};deleted!=true;"
+            f"source=={args.source}",
```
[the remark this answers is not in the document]
BorekZnovustvoritel: The important in this case is usage of the script. Since the [source task param is not optional](https://github.com/redhat-openshift-ecosystem/operator-pipelines/blob/main/ansible/roles/operator-pipeline/templates/openshift/tasks/reserve-operator-name.yml#L13), this implementation won't cause the declared problem.

BorekZnovustvoritel:
Seems like Preflight is slow today. I will move this to review, since the ISV test is the one relevant for this task.

mantomas (commented):
Looks good to me, just a clarifying question: the Slack discussion resulted in **exception for specific operator**, but this change, as I read it, will stop checking the uniqueness for all operators across sources. Is this the desired outcome? I see the AC is clear, but there was no grooming/discussion so I need to make sure we understand it the same way.

BorekZnovustvoritel:
Yes, I consulted this with Aleš in the office and he would choose the approach I used.

In the Openshift catalog, there is always an UI label showing from which source each operator comes. So this should be alright.

We can definitely discuss this on a standup in case you prefer a different approach.

BorekZnovustvoritel:
Based on the conversation, I changed the ticket and reworked the solution.

The working exception is visible in this PR:
https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/4038

From the logs it is obvious that the query for source is used there.

Successful pipeline without the exception is visible here:
https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/4034

operatorcert/entrypoints/reserve_operator_name.py
mantomas: Is this query out of scope now after the refactor? The comment on line 64 is about single package, but for operators with exception the API returns multiple results IIUC.

BorekZnovustvoritel: I think you are right and it's best to make the both queries behave the same. Updated.

BorekZnovustvoritel:
Manual retest PR with the exception added: https://github.com/redhat-openshift-ecosystem/operator-pipelines-test/pull/4039

mantomas (approved):

tomasbakk (approved):
lgtm 👍

